PR #31, opened 2019-08-08
Fix grammar added links

AlasdairGray:
Fixing a few minor grammatical issues in the text and adding links to concepts that may not be clear to everyone.


stain:
Thanks!

PR #35, opened 2019-08-09
Add list of publications, talks, citations and events

stain:
See preview at https://github.com/ResearchObject/ro-crate/blob/publications/docs/outreach.md


eocarragain:
LGTM.

PR #37, opened 2019-08-09
Describe RO-Crate metadata file

stain:
If there are nested directories included with `type: Dataset` it can be difficult to work out which `Dataset` is the _Root Data Entity_ - as developers are notoriously dubious at handling relative paths.

Therefore I suggest to add a new contextual item [RO-Crate metadata file](https://github.com/ResearchObject/ro-crate/blob/many-datasets/docs/0.2-DRAFT/index.md#ro-crate-metadata-file) which MUST be present if there are more than one `Dataset` mentioned. This is also a convenient place to describe the version of the _RO-Crate JSON_ that is implemented, as well as metadata provenance in case the RO-Crate JSON was converted from a Calcyte spreadsheet.

```json
        {
            "@type": [
                "CreativeWork"
            ],
            "about": {
                "@id": "."
            },
            "additionalType": {
                "@id": "https://w3id.org/ro/crate/0.2-DRAFT/"
            },
            "creator": {
                "@id": "https://orcid.org/0000-0001-9842-9718"
            },
            "isBasedOn": {
                "@id": "./CATALOG.xlsx"
            }
        }
```

---

I did not include a phrase about separate `license` although there could be a case for this, say Dataset is CC-BY but the metadata about the Dataset is CC0. This should instead be indicated with <http://schema.org/sdLicense> on the root dataset and/or the individual described items.


ptsefton:
Suggest this should have a fixed `identifier` to make it easy to locate the root dataset.

We can't rely on '@id's as they may change during processing, as they are resolved relative to the document base.

```
{
           "identifier": "ro-crate-metadata.jsonld",
            "@type": [
                "CreativeWork"
            ],
            "about": {
                "@id": "."
            },
            "additionalType": {
                "@id": "https://w3id.org/ro/crate/0.2-DRAFT/"
            },
            "creator": {
                "@id": "https://orcid.org/0000-0001-9842-9718"
            },
            "isBasedOn": {
                "@id": "./CATALOG.xlsx"
            }
        }

```

eocarragain:
Would https://schema.org/schemaVersion be better here than additionalType?

> Indicates (by URL or string) a particular version of a schema used in some CreativeWork. For example, a document could declare a schemaVersion using an URL such as http://schema.org/version/2.0/ if precise indication of schema version was required by some application.

stain:
Already solved by #42

PR #43, opened 2019-09-05
Tweaked draft reference and added more workflow info

alaninmcr:


eocarragain:
@stain can you take a look at this from a workflow perspective?

PR #48, opened 2019-10-23
Remove duplicate JSON-LD keys

stain:
Raised by @fbacall - we had some duplicate keys for some reason. This made the schema technically not valid JSON in some parson.


stain:
Although this change would update 0.2 context.json it is not adding any semantic change as the duplicate keys had the same values.

Issue #38, opened 2019-08-09
additionalType needs to be namespaced


stain:
**Describe the bug**

Adding `WorkflowSketch` etc to the `@context` is nice, but does not actually work with `additionalType` as the `@id` is interpreted in JSON-LD as a relative path.

**URL**
https://researchobject.github.io/ro-crate/0.2-DRAFT/#workflows-and-scripts

**Suggested fix**

Either we have to use a prefix so that it becomes expanded by the `@context` as the earlier `wfdesc:Workflow` (which exposes the different workflows), we use full URL, or we move `Workflow`, `Script` and `WorkflowSketch` to `@type: [array]` which always expand by the `@context` instead of relative path.

**Additional context**

```json
{  "@context": ["https://w3id.org/ro/crate/0.2-DRAFT/context", {"@base": "http://example.com/crate-1/"}],
            "@id": "workflow/retropath.knime",
            "@type": "SoftwareSourceCode",
            "additionalType": {"@id": "Workflow"},
            "name": "RetroPath Knime workflow",
            "description": "KNIME implementation of RetroPath2.0 workflow",
            "creator": {"@id": "#thomas"},
            "programmingLanguage": {"@id": "#knime"},
            "license": "https://spdx.org/licenses/BSD-2-Clause.html",
            "potentialAction": {
                "@type": "ActivateAction",
                "instrument": {"@id": "#knime"}
            }
        }

```

gives the triples in [JSON-LD playground](https://json-ld.org/playground/)

Subject | Predicate | Object | Language | Datatype | Graph
-- | -- | -- | -- | -- | --
_:b0 | http://schema.org/instrument | http://example.com/crate-1/#knime |   |   |  
_:b0 | http://www.w3.org/1999/02/22-rdf-syntax-ns#type | http://schema.org/ActivateAction |   |   |  
http://example.com/crate-1/workflow/retropath.knime | http://schema.org/additionalType | **http://example.com/crate-1/Workflow** |   |   |  
http://example.com/crate-1/workflow/retropath.knime | http://schema.org/creator | http://example.com/crate-1/#thomas |   |   |  
http://example.com/crate-1/workflow/retropath.knime | http://schema.org/description | KNIME implementation of RetroPath2.0 workflow |   |   |  
http://example.com/crate-1/workflow/retropath.knime | http://schema.org/license | https://spdx.org/licenses/BSD-2-Clause.html |   |   |  
[...]

PR #49, opened 2019-10-23
Tidy up Workflow/Script section

stain:
I modulated the table as well as I thought it was a bit too permissive.

Rendered on https://github.com/ResearchObject/ro-crate/blob/workflow-tidy/docs/0.3-DRAFT/index.md#workflows-and-scripts


stain:
This extends #43 - perhaps @alaninmcr  could review?

stain:
#55 points out `author` vs `creator` - we should not need both for workflows. The rest of RO-Crate use just `author` (although I know @ptsefton have found issues with contributor roles for other resources)

stain:
Agreed in call 2019-10-24 that the MUST/SHOULD in this table should be trimmed down for 1.0 release so we are not painting ourselves in a corner and require unachievable metadata.

stain:
Agreed in call 2019-10-24 that the #38 `additionalTypes` should be replaced with a `@type: []` multiple types array instead.

stain:
OK, I have trimmed down properties, simplified example, changed to `@type` array, removed `creator`. 

Notably I also removed `encodingFormat` as I think that is given by `programmingLanguage` - you may disagree as that is less consistant with than say `encodingFormat: "text/x-python"` - opinions?

I also changed one example to show a Python script.

stain:
@ptsefton accepts

PR #56, opened 2019-10-31
Schema version

stain:
Use http://schema.org/schemaVersion instead of http://schema.org/identifier to mark the RO-Crate Metadata File Descriptor:

```json

{ "@context": "https://w3id.org/ro/crate/0.3-DRAFT/context", 
  "@graph": [
    {
        "@type": "CreativeWork",
        "schemaVersion": "https://w3id.org/ro/crate/0.3-DRAFT",
        "about": {"@id": "./"}
    },
    
    {
      "@id": "./",
      "@type": "Dataset",
      ...
    }
  ]
}
```

I think this can also be a nice way to show the version of RO-Crate that we try to adhere to semantically. It is separate to the `@context` and as a string survives any JSON-LD processing and is easier to compare than say `additionalType: {"@id": "https://w3id.org/ro/crate/0.3-DRAFT"}`.


stain:
Perhaps @danbri can comment on if this is inappropriate use of https://schema.org/schemaVersion !

danbri:
schemaVersion was more designed for the use case of a conservative (lowercase-c) publisher wanting to be explicit about which version of schema.org they were intending. We didn't want to put version numbers in the URLs of the vocabulary but if someone *did* want to say "we mean v3.0 version of http://schema.org/Person", they could use this property against the document using the markup.

jmfernandez:
> schemaVersion was more designed for the use case of a conservative (lowercase-c) publisher wanting to be explicit about which version of schema.org they were intending. We didn't want to put version numbers in the URLs of the vocabulary but if someone _did_ want to say "we mean v3.0 version of http://schema.org/Person", they could use this property against the document using the markup.

@danbri So, `schemaVersion` was thought to play a equivalent role to **versionIRI** predicate in OWL, right?

danbri:
It looks broadly comparable, but schema.org/schemaVersion accepts textual
values too

stain:
Thanks, @danbri - my initial understanding was that `schemaVersion` could relate to the abstract  "schema version" that could include extensions such as ours, but you seem to imply that if we provide `schemaVersion` then it should really relate to the version of the schema.org vocabulary only? 

In this case would it be right to instead have the  `"schemaVersion": {"@id": "http://schema.org/version/5.0/"} - for archival purposes that would be the schema.org baseline the RO-Crate have interpreted (although with a different `@context`). 

We should probably indicate the schema.org version anyway in the versioned RO-Crate docs and/or our derived context - I think properties and classes are not removed, just deprecated, but it would show which underlying textual definitions we refer to, as they change more over time.

stain:
Assuming multiple  `schemaVersion` values would be confusing here, if we want to indicate which particular interpretation/profile of schema.org we use (e.g. https://w3id.org/ro/crate/0.2), then if that can't be set with `schemaVersion` we could perhaps use another property, say http://purl.org/dc/terms/conformsTo

I think https://schema.org/encodingFormat could be misleading at that should be `"application/ld+json"` for the encoding of the file data, rather than the conventions followed.

To confuse matters the [JSON-LD IANA media type](https://www.w3.org/TR/json-ld/#iana-considerations) has a `profile` parameter, which can be used to indicate flattened, compacted and other profiles. In https://researchobject.github.io/ro-crate/0.3-DRAFT/#ro-crate-json-ld-media-type I suggested  we use it like this if serving `ro-crate-metadata.jsonld` over HTTP:

```http
    Content-Type: application/ld+json; profile="http://www.w3.org/ns/json-ld#flattened 
                   http://www.w3.org/ns/json-ld#compacted https://w3id.org/ro/crate"
```

stain:
Agreed to add `conformsTo` to context and use it for this purpose

stain:
OK, I added a new script [schema-context.py](https://github.com/ResearchObject/ro-crate/blob/schemaVersion/scripts/schema-context.py) to re-generate our context.json using [schema.org context 5.0](http://schema.org/version/5.0/) and adding our mapped non-schema.org terms.  

Using `ident=5` you see in the diff there were some, but not many changes!

btw, I think `context.json` on `1.0` release should be renamed `context.jsonld` to get the correct `Content-Type` from Github Pages (it works). I won't rename now as I would have to update w3id mapping.

stain:
Note, 0.2 mistakenly had lowercase `journal` mapped to http://schema.org/Periodical - this now match the HTML which says camel-case `Journal`

PR #57, opened 2019-11-01
Pedantic mode

stain:
Working with @fbacall we found various issues with `@id` and `identifier`. Some of the examples were outdated.


stain:
Agreed that `./foo` should be `foo`

Still to be investigated `foo/` vs `foo` for directories

Blank nodes ARE permitted, modify "MUST URI" to "SHOULD URI". `#` ones are OK, but also `_:` ones. `#` example changes OK.

stain:
OK, so the remaining contentious part here is directories should end in `/` or not and if we should say something about it. I'll raised that as new issue.

Issue #60, opened 2019-11-11
Tables for MUST, SHOULD etc in  spec are too hard to complete


ptsefton:
The MUST / SHOULD tables are too complicated to complete in time for a release.

Some reasons:
*  The Data citation use case is not well specified and will require development to inter-operate with DataCite software
*  The Google Dataset search integration is untested - we don't have data to advise people of what to put in RO-Crates.
*  The institutional repository case will vary by institution so we can't do that as part of the spec.

**Suggested fix**

I suggest we remove the tables and replace with a simple list of MUST have elements for a minimal RO-Crate - then start working on Implementation guides for people who want to be DataCite or Google Dataset search compliant.

My Suggestion for MUST:  

-  @type MUST be (at least) schema:Dataset 
-  @id Must be a a string of ‘./’  (with the  _RO-Crate Metadata File Descriptor_  present as well) 
-  name - So humans can identify the crate
-  datePublished  To tell different versions apart
-  license - so that people know what they are allowed to do with the data including - 
-  systems administrators and people are cleaning up after the creator has gone
contactPoint so there is an email address.

A future version of this specification will have implementation guides for data publishers which explain how to create citable datasets.

The following properties are SHOULDs - to assist in distributing data
-  author  SHOULD  appear at least once | referencing a Contextual Entity of @type: Person or Organization 

And where there is publishing support available:
-  identifier - Such as DOI, or a URL - but this depends on repository infrastructure 
-  distribution Link to a downloadable version of the data

ptsefton:
@stain What do you think? Can't see us getting those tables sorted in the next couple of days.

AlasdairGray:
To be compliant with Google dataset search minimal requirements, then `schema:description` is a MUST property. (They currently use name and description only.)

stain:
Shall we say that name and description are MUST on dataset, but SHOULD on the other types? Is it valid to always require a description? I found the schema.org structured data testing tool also has opinions on the LENGTH of the description (!)

Then the rest of the "?" can be MAY for now (meaning they are just informationally listed examples from schema.org). Future RO-Crate 1.1 can then upgrade them to SHOULD (but not to MUST).

--
Stian Soiland-Reyes
The University of Manchester
https://www.esciencelab.org.uk/
https://orcid.org/0000-0001-9842-9718

On 11 Nov 2019 10:20, Alasdair Gray <notifications@github.com> wrote:

[...]

PR #61, opened 2019-11-11
Tidyup

ptsefton:
@stain Edits as discussed this deals with most of the outstanding stuff I think but we still need to deal with the tables of MUSTs etc. see #60

### DIFF
--- a/docs/0.3-DRAFT/index.md
+++ b/docs/0.3-DRAFT/index.md
@@ -828,7 +828,6 @@ A RO-Crate SHOULD have contact information, using a [contactPoint](http://schema
 
 NOTE: This usage follows that of Google Dataset search.
 
-IMPLEMENTATION NOTE: The Google structured Data tool gives an error if `mailto:` URLs are used in the "url" property.
 
 ```json
 {
--- a/docs/0.3-DRAFT/index.md
+++ b/docs/0.3-DRAFT/index.md
@@ -79,7 +79,7 @@ _RO-Crate Metadata File_: A JSON-LD file stored as `ro-crate-metadata.jsonld` in
 
 _RO-Crate Website_: Human-readable HTML pages which describe the RO-Crate (i.e. the _Root Data Entity_, its _Data Entities_ and _Context Entities_), with a home-page at `ro-crate-preview.html` (any additional files reside in `ro-crate-preview_files/`)
 
-_Data Entity_: A JSON-LD representation, in the _RO-Crate Metadata File_, of a directory, file or other resource contained or described by the RO-Crate.
+_Data Entity_: A JSON-LD representation, in the _RO-Crate Metadata File_, of a directory or file or other resource contained in the RO-Crate.
 
 _Root Data Entity_: A _Data Entity_ of type [Dataset], representing the RO-Crate as a whole.  
 
@@ -116,20 +116,21 @@ The structure an _RO-Crate_ MUST follow is:
 
 ```
 <RO-Crate root directory>/
-|   ro-crate-metadata.jsonld  # RO-Crate Metadata File MUST be present
-|   ro-crate-preview.html     # RO-Crate Website homepage MAY be present
-|   ro-crate-preview_files/   # MAY be present
-|    | [other RO-Crate Website files]
-|   [payload files and directories]  # 1 or more SHOULD be present
+  |   ro-crate-metadata.jsonld  # _RO-Crate Metadata File_ MUST be present
+  |   ro-crate-preview.html     # _RO-Crate Website_ homepage MAY be present
+  |   ro-crate-preview_files/   # MAY be present
+  |    | [_RO-Crate Website_ files]
+  |   [payload files and directories]  # 1 or more SHOULD be present
 ```
 
-The name of the _RO-Crate root_ directory is not defined, but a root directory is identifiable by the presence of `ro-crate-metadata.jsonld`. For instance, if an _RO-Crate_ is archived in a ZIP-file, the ZIP root directory correspond to _RO-Crate root_ directory if it contains `ro-crate-metadata.jsonld`.
+The name of the _RO-Crate root_ directory is not defined, but a root directory is identifiable by the presence of `ro-crate-metadata.jsonld`. For instance, if an _RO-Crate_ is archived in a ZIP-file, the ZIP root directory corresponds to _RO-Crate root_ directory if it contains `ro-crate-metadata.jsonld`.
 
 <!--
 It follows that RO-Crates can be _nested_ by including payload directories that themselves contain `ro-crate-metadata.jsonld`.
 -->
 
-### RO-Crate Metadata File (`ro-crate-metadata.jsonld`)
+
+### _RO-Crate Metadata File_ (`ro-crate-metadata.jsonld`)
 
 *  The _RO-Crate Metadata File_ MUST be named `ro-crate-metadata.jsonld` and appear in the _RO-Crate Root_
 * The _RO-Crate Metadata File_ MUST contain _RO-Crate JSON-LD_; a valid [JSON-LD 1.0](https://www.w3.org/TR/2014/REC-json-ld-20140116/) document in [flattened](https://www.w3.org/TR/json-ld/#flattened-document-form) and [compacted](https://www.w3.org/TR/json-ld/#compacted-document-form) form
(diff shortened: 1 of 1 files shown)


docs/0.3-DRAFT/index.md
```diff
@@ -116,20 +192,21 @@ The structure an _RO-Crate_ MUST follow is:
 
 ```
 <RO-Crate root directory>/
-|   ro-crate-metadata.jsonld  # RO-Crate Metadata File MUST be present
-|   ro-crate-preview.html     # RO-Crate Website homepage MAY be present
-|   ro-crate-preview_files/   # MAY be present
-|    | [other RO-Crate Website files]
-|   [payload files and directories]  # 1 or more SHOULD be present
+  |   ro-crate-metadata.jsonld  # _RO-Crate Metadata File_ MUST be present
```
stain: This `_italics_` does not work inside a "```" pre block. We would have to rewrite it to use "\`foo\`" on each line or pure HTML :/

ptsefton: The  italics can go. In the next update I've moved this to the end with the bit about repo-specific IDs to make the whole thing a bit less dense

docs/0.3-DRAFT/index.md
```diff
@@ -346,11 +383,108 @@ start with `https://w3id.org/ro/crate/`.
...
+If possible, _Contextual Identifiers_ SHOULD have URIs as values for their `@id`:
+*   For People participating in the research process: [ORCID](https://orcid.org) identifiers.
+*   For [Organization](http://schema.org/Organization)s including [funder](http://schema.org/funder)s,  Research Organization Registry URIs. [https://ror.org/](https://ror.org/)
+*   For items of type [Place], a geonames URL.
+*   For file formats, a [Pronom] URL, for example <https://www.nationalarchives.gov.uk/PRONOM/fmt/831>.
+*   In other cases, a stable, useful URL that resolves to something helpful to describe the entity.
+
+If there is no usable URI to identify an entity, _Contextual Entities_ MAY have local IDs (prefaced with a `#` or JOSN blank node identifiers prefaced with `_:`.
+
+Identifiers for _Contextual Entities_ MUST NOT use `@id`s which can be construed as paths. So `#NikonCamera1` is acceptable but `NikonCamera1` is NOT.
```
stain: I think lower this to SHOULD - it would be valid - particularly on the web, that the contextual entity has a direct resource representation (which is not a `File`).  You have earlier mentioned instruments with wiki pages, I don't see why you could not do that here as well, but you would need to know what you are doing (hence thoughtfully violate a SHOULD)

docs/0.3-DRAFT/index.md
```diff
@@ -363,11 +497,6 @@ types of dataset may put further constraints or requirements of metadata beyond
 the Root Data Entity (see Extending RO-Crate below).
 
 
-The _RO-Crate Metadata File Descriptor_ MAY contain information such as
```
stain: OK to remove, this is mentioned elsewhere.

docs/0.3-DRAFT/index.md
```diff
@@ -588,7 +717,7 @@ If there is no web-accessible description for a file format it SHOULD be describ
     "@type": "File",
     "contentSize": "120",
     "description": "A file in a non-standard format",
-    "encodingFormat": ["text/plain", {"@id": "https://www.commonwl.org/v1.0/Workflow.html"}]
+    "encodingFormat": ["text/plain", {"@id": "some_extension.md"}]
```
stain: Good correction - as indeed CWL has version of its own language as REQUIRED in every `*.cwl` file ;) 

```yaml
#!/usr/bin/env cwl-runner
cwlVersion: v1.0
class: Workflow
```

which is documented at permalink <https://w3id.org/cwl/v1.0/> which then would be a natural `@id` to use for `encodingFormat` (shown in Workflow section I think)

docs/0.3-DRAFT/index.md
```diff
@@ -1324,7 +1276,7 @@ To record curation actions which modify a [File](http://schema.org/MediaObject)
 ```json
 {
     "@id": "#history-01",
-    "@type": "CreateAction",
+    "@type": "#CreateAction",
```
stain: No `#`, those types are schema.org terms from the `@context`, e.g. <https://schema.org/CreateAction>

Still there COULD be other types of custom actions like `#ImagineAction`, but perhaps we should not go into that.

docs/0.3-DRAFT/index.md
```diff
@@ -1378,7 +1328,90 @@ To record curation actions which modify a [File](http://schema.org/MediaObject)
...
+### Workflows and scripts
+
+Scientific workflows and scripts that can be or were used to analyze or generate files contained in an the RO-Crate MAY be embedded in an RO-Crate. Workflows and scripts are described using the type SoftwareSourceCode instead of `SoftwareApplication` above.  
+
+The distinction is fluid, and comes down to the availability and understandability of code to re-run a workflow. For instance, office spreadsheet applications are generally available and do not need further explanation; while a Python script that is customized for a particular data analysis might be important to understand further and SHOULD be included as source code in the RO-Crate dataset:
+
+```json
+       {
+            "@id": "workflow/retropath.knime",  
+            "@type": ["File", "SoftwareSourceCode", "Workflow"],
```
stain: Good for consistency to also include `File` here (although verboser!)

docs/0.3-DRAFT/index.md
```diff
@@ -1378,7 +1328,90 @@ To record curation actions which modify a [File](http://schema.org/MediaObject)
...
+            "programmingLanguage": {"@id": "https://www.knime.com/knime-software/knime-analytics-platform"},
+            "license": "https://spdx.org/licenses/BSD-2-Clause.html",
+            "potentialAction": {
+                "@type": "ActivateAction",
+                "instrument": {"@id": "https://www.knime.com/knime-software/knime-analytics-platform"}
+            }
+        }
+```
+
+Workflows and scripts are saved on disk using a language and generally need a runtime, in RO-Crate this SHOULD be indicated using a liberal interpretation of `programmingLanguage` and `potentialAction`. Note that the language and its runtime might differ, e.g. there are multiple potential runtimes of a JavaScript, although in the base-case they are the same, and as a shortcut can be described in one go as both a `ComputerLanguage` and `SoftwareApplication`:
```
stain: This text mentioning `potentialAction` and `alternateName` below seems to reflect the Workflow section **before** my merged `tidy-workflow` from #49  - now I have to check everything else in all the moved sections to see if they are also outdated

docs/0.3-DRAFT/index.md
```diff
@@ -1378,7 +1328,90 @@ To record curation actions which modify a [File](http://schema.org/MediaObject)
...
+                "SoftwareApplication"
+            ],
+            "name": "KNIME Analytics Platform",
+            "alternateName": "KNIME",
+            "url": {"@id": "https://www.knime.com/knime-software/knime-analytics-platform"},
+            "version": "3.6"
+        }
+```
+
+A data entity representing a runtime application or language MUST have a [name], [url] and [version], which SHOULD indicate a known version the workflow/script was developed or tested with. [alternateName] MAY be provided if there is a shorter colloquial name, for instance “R” instead of “The R Project for Statistical Computing”.
```
stain: This text may be outdated

docs/0.3-DRAFT/index.md
```diff
@@ -1378,7 +1328,90 @@ To record curation actions which modify a [File](http://schema.org/MediaObject)
...
+
+To indicate that a `SoftwareSourceCode` is in the form of a script use a `@type` of `WorkflowScript`. If the `SoftwareSourceCode` is in the form of a workflow (e.g. a pipeline of steps with data flow) `Workflow`.
+
+#### Workflow diagram/sketch
+
+It is useful to show a diagram or sketch to explain the script/workflow. This may have been generated from the workflow management system or drawn manually as a diagram. It SHOULD be included as an `ImageObject` which is `about` the `SoftwareSourceCode`:
```
stain: What do you think, should we keep this as `about` from the image to the workflow, or [`image`](http://schema.org/image) **from** the workflow to the image?  The image may not have been generated programmatically from the workflow file, but often is.

docs/0.3-DRAFT/index.md
```diff
@@ -1378,7 +1328,90 @@ To record curation actions which modify a [File](http://schema.org/MediaObject)
...
+            "@id": "workflow/workflow.svg",
+            "@type": ["ImageObject","WorkflowSketch"],
+            "encodingFormat":  "image/svg+xml",
+            "description": "Diagram of RetroPath2.0 workflow",
+            "about": {"@id": "workflow/workflow.knime"}
+}
+```
+
+
+The image file format SHOULD be indicated with [encodingFormat] via Mime - eg  "image/svg+xml".
```
stain: "Media type", not "Mime" (which has to do with email)

docs/0.3-DRAFT/index.md
```diff
@@ -1388,32 +1421,33 @@ To include EXIF, or other data which can be encoded as property/value pairs, add
...
+  ]
+}
 
 {
-      "@id": "#c2521494-9b94-4b23-a713-6b281f540823",
-      "@type": "PropertyValue",
-      "name": "InternalSerialNumber",
-      "value": "4102011002108002               "
-    },
+  "@id": "_:b312",
```
stain: Yes, `_:` identifiers now allowed again, and they very much make sense here.

docs/0.3-DRAFT/index.md
```diff
@@ -1586,7 +1620,9 @@ For example, this data is exported from an [Omeka](https://omeka.org) repository
 }
 ```
 
+[See an example of a repository exported to a RO-Crate.](https://data.research.uts.edu.au/examples/v1.0/farms_to_freeways/)
```
stain: Old text? I thought you agreed we could not link to these yet without stable URIs? Example is also outdated datacrate 1.0 rather than RO-Crate.

docs/0.3-DRAFT/index.md
```diff
@@ -176,9 +253,9 @@ If present in the root directory, `ro-crate-preview.html` MUST:
   ```
 
 `ro-crate-preview.html` SHOULD:
-*   Contain at least the same information as the _RO-Crate JSON-LD_, with the exception that files which have no description, author or similar metadata MAY not be listed in the website.
-*   Display at least the metadata relating to the _Root Data Enity_ as static HTML without the need for scripting. It MAY contain extra features enabled by JavaScript.
-*   When a _Data Entity_ or _Contextual Entity_ is referenced by its ID:
+*   Contain a human-readable summary of information as the _RO-Crate JSON-LD_, with the exception that files which have no description, author or similar metadata MAY not be listed in the website.
+*   Display at least the metadata relating to the _Root Data Enity_ as static HTML without the need for JavaScript being enabled. It MAY contain extra features enabled by JavaScript.
+*   When a _Data Entity_ or _Context Entity_ is referenced by its ID:
```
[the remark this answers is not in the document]
ptsefton: Fixed on 3b5f705

PR #62, opened 2019-11-12
Redoing edits

ptsefton:
I redid yesterday's mess based on the master branch. No major re-ordering in this so it should merge OK. Mostly tidying up, but also removes the MUST have tables as per #60 .

Made some changes in the workflow section to make the examples more consistent.

Added @type File (as an array element) to a lot of Data Entities that did not have it (eg Workflow examples that are actually describing files) - I think this would make it easier for coders but maybe we should discuss this.

Hope this is cleaner than yesterday's @stain!


stain:
Very nice - I did the section move separately in c3e643cec6fcb3e7f96b6e300a826ff29dee60d9

PR #73, opened 2020-04-08
Avoid @base: null as it breaks parsing as RDF

stain:
Avoid using `@base: null` as it means parsing RO-Crate JSON-LD as RDF will skip triples with relative URIs. 

Rather adding `@base: null` can be recommended for the compaction algorithm if making RO-Crates.

### DIFF
--- a/docs/1.1-DRAFT/context.jsonld
+++ b/docs/1.1-DRAFT/context.jsonld
@@ -2492,6 +2492,5 @@
           "wfprov": "http://purl.org/ro/wfprov#",
           "roterms": "http://purl.org/ro/roterms#",
           "wf4ever": "http://purl.org/ro/wf4ever#",
-          "@base": null
      }
 }
--- a/docs/1.1-DRAFT/context.jsonld
+++ b/docs/1.1-DRAFT/context.jsonld
@@ -2491,6 +2491,6 @@
           "wfdesc": "http://purl.org/ro/wfdesc#",
           "wfprov": "http://purl.org/ro/wfprov#",
           "roterms": "http://purl.org/ro/roterms#",
-          "wf4ever": "http://purl.org/ro/wf4ever#",
+          "wf4ever": "http://purl.org/ro/wf4ever#"
      }
 }
--- a/docs/1.1-DRAFT/index.md
+++ b/docs/1.1-DRAFT/index.md
@@ -2272,6 +2272,7 @@ Parsing this as RDF will generate triples including:
 
 Here consumers can assume `/` is the _RO-Crate Root_ and generating relative URIs can safely be achieved by  search-replace as the arcp URI is unique. Saving _RO-Crate JSON-LD_ from the triples can be done by using the arcp URI to [relativize absolute URIs within RO-Crate Root](#relativizing-absolute-uris-within-ro-crate-root).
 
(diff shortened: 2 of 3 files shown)


stain:
Example of it breaking [in JSON-LD Playground](https://tinyurl.com/qrjn6m3) - select N-Quads and no triples appear.

with [this context](https://rawcdn.githack.com/ResearchObject/ro-crate/8c85e590155d0e81d47786628838ab25a1035359/docs/1.1-DRAFT/context.jsonld) it gives [all the triples](https://tinyurl.com/ufo266b) back under N-Quads.

stain:
As you see in Playground example the RDF would easily then get the wrong base URI like <https://json-ld.org/playground/lots_of_little_files/> depending on how the JSON-LD is parsed. So we would need a paragraph about the `@context: null` hack.

stain:
## Explainer

(at least according to why I think this happened)

In RDF documents (except N-Triples and N-Quads) RDF subjects, predicates and objects can be given as a _IRI reference_ - that is a partial URI that is resolved according to the _base URI_ of the document, to form the absolute URI.

For instance, if `http://example.com/crates/15/ro-crate-manifest.jsonld` is the document, and it contained a statement about `./` it is resolved to the absolute URI `http://example.com/crates/15/` - and similar for `..` and `nested/foo.txt` etc.

This is important because in a loaded _RDF Graph_ the subject/predicate/object must be _absolute URIs_ - for instance because you want to compare two RDF Graphs loaded from different base documents. 

Now in JSON-LD contexts one can set `@base: null` to force the URIs to remain relative. This means that in the JSON-LD document one would still find `./` and `nested/foo.txt` etc. This is OK, as a JSON-LD document is a _serialization_ of an RDF Graph.  However if you then use the same JSON-LD libraries to get the RDF graph, triples involving non-absolute URIs will be skipped.

So I think we need to recommend adding `@base: null` only for use when compacting/flattening a JSON-LD to be saved back out, but can't keep that in our `@context` as that would ruin RDF parsing for everyone (it's much harder to calculate manually and inject in the `@base` of a remote web document)

stain:
This was discussed in [RO-Crate community call 2020-05-28](https://s.apache.org/ro-crate-minutes#heading=h.w3n6gm7wo3vg) and agreed to remove `@base: null` to allow RDF parsing into absolute URIs, but to also add advice on how JSON-LD processing/compacting should use `@base:null` to ensure URIs stay relative when saving the JSON-LD.

@markwilkinson agreed to comment on this issue to add log/example of how RDF parsing of relative URIs with RO-Crate 1.0 context failed or lost triples.

stain:
Added [appendix subsection](https://github.com/ResearchObject/ro-crate/commit/d2d3b6ecc7898480d255aba619ffa30cce1761fb?diff=unified&short_path=ee90865#diff-ee90865049d77cbc35bed8c750ba446e) on dealing with `@base` when JSON-LD flattening, JSON-LD Expanding and parsing as RDF.

@markwilkinson would you review the RDF triple part of it?

@ptsefton - should I also add something about compacting/framing to the _compact_ tree JSON, or is this implied by the Expansion section?

stain:
Thanks, @markwilkinson I'll consider this PR as passed review and merge :)

PR #74, opened 2020-04-08
Adding @id for external URIs

stain:
Modifying 1.1-DRAFT spec to allow `@id` of Data Entity to be external/absolute URIs, and adds considerations before adding external URI to `hasPart` of an RO-Crate.

See #72 .

### DIFF
--- a/docs/1.1-DRAFT/index.md
+++ b/docs/1.1-DRAFT/index.md
@@ -606,8 +606,8 @@ Example of an RO-Crate where some of the _File Data Entities_ are external to th
       {
         "@id": "https://zenodo.org/record/3541888/files/ro-crate-1.0.0.pdf"
       },
-      ],
-   },
+      ]
+  },
   {
     "@id": "survey-responses-2019.csv",
     "@type": "File",
--- a/docs/1.1-DRAFT/index.md
+++ b/docs/1.1-DRAFT/index.md
@@ -126,13 +126,15 @@ The structure an _RO-Crate_ MUST follow is:
 |   ro-crate-preview.html     # RO-Crate Website homepage MAY be present
 |   ro-crate-preview_files/   # MAY be present
 |    | [other RO-Crate Website files]
-|   [payload files and directories]  # 1 or more SHOULD be present
+|   [payload files and directories]  # 0 or more
 ```
 
 The name of the _RO-Crate root_ directory is not defined, but a root directory is identifiable by the presence of `ro-crate-metadata.jsonld`. For instance, if an _RO-Crate_ is archived in a ZIP-file, the ZIP root directory correspond to _RO-Crate root_ directory if it contains `ro-crate-metadata.jsonld`.
 
+[Data Entities](#core-metadata-for-data-entities) in the RO-Crate MUST either be _payload files/directories_ present within the RO-Crate root directory or its subdirectories, or be [Web-based Data Entities](#web-based-data-entities).
+
 <!--
-It follows that RO-Crates can be _nested_ by including payload directories that themselves contain `ro-crate-metadata.jsonld`.
+RO-Crates can be _nested_ by including payload directories that themselves contain `ro-crate-metadata.jsonld`.
 -->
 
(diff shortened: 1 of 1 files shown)


docs/1.1-DRAFT/index.md
```diff
@@ -377,14 +379,13 @@ The _RO-Crate Metadata File Descriptor_ MAY contain information such as
...
 
-
 The table below outlines the properties that the _Root Data Entity_ MUST have to be minimally valid and additionally highlights properties required to meet other common use-cases, including the minimum metadata necessary to mint a DataCite DOI:
 
 #### Direct properties of the Root Data Entity
 
 The _Root Data Entity_ MUST have the following properties:
 *  `@type`: MUST be [Dataset]
-*  `@id`:  MUST be a a string of ‘./’
+*  `@id`:  MUST end with `/` and SHOULD be the string `./`
```
fbacall, line 388: How can a parser know what is the _Root Data Entity_ in this case?

stain: They would need to follow the link from the [RO-Crate Metadata File Descriptor ](https://researchobject.github.io/ro-crate/1.0/#core-metadata-for-the-root-data-entity)

> 1. For each entity in `@graph` array
> 2. ..if the `conformsTo` property is a URI that starts with `https://w3id.org/ro/crate/`
> 3. ....from this entity's `about` object keep the `@id` URI as variable _root_
> 4. For each entity in `@graph` array
> 5. .. if the entity has an `@id` URI that matches _root_ return it

but I am also not too confident with this change - perhaps @markwilkinson can comment on if it is still needed to have absolute URIs as in <http://ldp.cbgp.upm.es:8890/DAV/coronavirus/ro-crates/GobiernoEspana/ro-crate-metadata.jsonld> ?

```jsonld
    {
      "@type": "CreativeWork",
      "about": {
        "@id": "http://w3id.org/FAIR_COVID19/DAV/coronavirus/ro-crates/GobiernoEspana"
      }
  {
      "@id": "http://w3id.org/FAIR_COVID19/DAV/coronavirus/ro-crates/GobiernoEspana",
      "@type": "Dataset",
      "name": [
        "Situación de COVID-19 en España"
      ]
}

Comment (fbacall) [reply docs/1.1-DRAFT/index.md:388]: I think he is using my library, and that property is not included!

fbacall/ro-crate-ruby#12

**Comment by fbacall** [comment]
When fetching a remote data entity, where should the local file be stored?

My assumption would be at the root of the crate, but what if it needs to go under `data/input/foo/bar`?

**Comment by stain** [comment]
@ptsefton you argued that remote files should be indicated as `RepositoryObject`s and not `File`?

Would you be able to sketch out how a modified example from [digital-library-and-repository-content section](https://researchobject.github.io/ro-crate/1.0/#digital-library-and-repository-content) would look?

@fbacall in https://github.com/ResearchObject/ro-crate/pull/74#issuecomment-627850469 You seem to be walking into the world of [BagIt fetch.txt](https://www.rfc-editor.org/rfc/rfc8493.html#section-2.2.3) - in which case I would expect the RO-Crate to refer to `@id: data/bigfile.txt` and `fetch.txt` say where `data/bigfile.txt` could be downloaded from.

**Comment by ThomasThelen** [comment]
In regards to establishing a link between external and local data...

What about using `schema:sameAs` on the `File` object?

```
  {
    "@id": "survey-responses-2019.csv",
    "@type": "File",
    "contentSize": "26452",
    "description": "Survey responses",
    "encodingFormat": "text/csv",
    "sameAs": "https://zenodo.org/record/3541888/files/ro-crate-1.0.0.pdf"
  },
  {
    "@id": "https://zenodo.org/record/3541888/files/ro-crate-1.0.0.pdf",
    "@type": "File",
    "contentSize": "310691",
    "description": "RO-Crate specification",
    "encodingFormat": "application/pdf",
  }
```

This is likely something that @ieguinoa would run across with https://github.com/ResearchObject/ro-crate-py/issues/2 while resolving remote data.

Issue #77, opened 2020-04-30
Doc update : make it clear that Dataset @id is a URIPath w/ *nix separators


marcolarosa:
When traversing a file path in code to create datasets or files, if on windows, the paths would come back with '\\' instead of '/'. It's not immediately obvious that when made into relative paths they should also be converted from '\\' to '/' in order to be valid URIPaths.

Could a note about this be made in the section "Core Metadata for Data Entities" just to make it clear that a windows path is *not* a valid URI path?

stain:
This is already described in section [Describing entities in JSON-LD](https://researchobject.github.io/ro-crate/1.0/#describing-entities-in-json-ld) in the appendix, so perhaps just add an internal link about that from the "Core Metadata for Data Entities" section?

stain:
Added below _Core Metadata for Data Entities_

#### Encoding file paths

Note that all `@id` [identifiers must be valid URI references](https://researchobject.github.io/ro-crate/1.0/#describing-entities-in-json-ld), care must be taken to express any relative paths using `/` separator, correct casing, and escape special characters like space (`%20`) and percent (`%25`), for instance a _File Data Entity_ from the Windows path `Results and Diagrams\almost-50%.png` becomes `"@id": "Results%20and%20Diagrams/almost-50%25.png"` in the _RO-Crate JSON-LD_.
 
In this document the term _URI_ includes international _IRI_s; the _RO-Crate Metadata File_ is always UTF-8 and international characters in identifiers SHOULD be written using native UTF-8 characters (_IRI_s), however traditional URL encoding of Unicode characters with `%` MAY appear in `@id` strings. Example: `"@id": "面试.mp4"` is preferred over the equivalent `"@id": "%E9%9D%A2%E8%AF%95.mp4"`

PR #80, opened 2020-06-03
Make it clear that @id is URIpath with / seperators

stain:
Added below _Core Metadata for Data Entities_ to clarify #77 

#### Encoding file paths

Note that all `@id` [identifiers must be valid URI references](https://researchobject.github.io/ro-crate/1.0/#describing-entities-in-json-ld), care must be taken to express any relative paths using `/` separator, correct casing, and escape special characters like space (`%20`) and percent (`%25`), for instance a _File Data Entity_ from the Windows path `Results and Diagrams\almost-50%.png` becomes `"@id": "Results%20and%20Diagrams/almost-50%25.png"` in the _RO-Crate JSON-LD_.
 
In this document the term _URI_ includes international _IRI_s; the _RO-Crate Metadata File_ is always UTF-8 and international characters in identifiers SHOULD be written using native UTF-8 characters (_IRI_s), however traditional URL encoding of Unicode characters with `%` MAY appear in `@id` strings. Example: `"@id": "面试.mp4"` is preferred over the equivalent `"@id": "%E9%9D%A2%E8%AF%95.mp4"`

### DIFF
--- a/docs/1.1-DRAFT/index.md
+++ b/docs/1.1-DRAFT/index.md
@@ -1,7 +1,7 @@
 <!--
-   Copyright 2019 University of Technology Sydney
-   Copyright 2019 The University of Manchester UK 
-   Copyright 2019 RO-Crate contributors <https://github.com/ResearchObject/ro-crate/graphs/contributors>
+   Copyright 2019-2020 University of Technology Sydney
+   Copyright 2019-2020 The University of Manchester UK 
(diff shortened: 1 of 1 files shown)


stain:
Would @marcolarosa or @ptsefton be able to review?

marcolarosa:
@stain Yep - that makes sense and it looks like that commit is putting that text into Core Metadata for Entities section which I think is also a good place for it.

PR #90, opened 2020-09-15
How to specify ad-hoc properties

### LINKED ISSUE #71: Use Case: As a researcher I want to be able to define ad hoc properties and use them in a crate without resorting to the verbose PropertyValue thing
I have discussed this in a [blog post](http://ptsefton.com/2020/03/24/semantic_web_hacks.htm).

I think the second solution I provide here may form the basis for a way to satisfy this Use Case.
 
In a nutshell:
-  You define an additional context term (Property) with a URL eg at example.org 
-   Add a definition of the Property to the @graph - and the RO-Crate HTML viewer can then use this to add a definition

stain:
Fixes #71

### DIFF
--- a/docs/1.1-DRAFT/index.md
+++ b/docs/1.1-DRAFT/index.md
@@ -2020,35 +2020,38 @@ Context terms must ultimately map to HTTP(s) URIs which poses challenges for cra
 
 RO-Crate provides some strategies To add a new term (a Class or Property) that is not in schema.org or another published vocabulary, so that there is a stable URI that can be added to the @context. 
 
-#### Choosing URLs for terms
+#### Choosing URLs for ad hoc terms
 
+For projects that have their own web-presence, URLs MAY defined and SHOULD resolve to useful content. For example for a project with web page <https://criminalcharacters.example.com/> the property `education` could have a URL: https://criminalcharacters.example.com/vocab#education which resolves to an HTML page that explains the term.
 
-For projects that have their own web-presence URLs MAY defined and SHOULD resolve to useful content. For example for a project the property `education` could have a URL: https://criminalcharacters.com/vocab/#education which resolves to an HTML end-point which explains the term.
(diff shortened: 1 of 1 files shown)


dgarijo:
Thanks Stian. I have to update a few guidelines and examples in the ro-terms repo, but this looks very good to me.

docs/1.1-DRAFT/index.md
```diff
@@ -2013,6 +2013,53 @@ When generating the _RO-Crate Website_ from _RO-Crate JSON-LD_, the code MUST us
...
+
+RO-Crate provides some strategies To add a new term (a Class or Property) that is not in schema.org or another published vocabulary, so that there is a stable URI that can be added to the @context. 
+
+#### Choosing URLs for ad hoc terms
+
+For projects that have their own web-presence, URLs MAY defined and SHOULD resolve to useful content. For example for a project with web page <https://criminalcharacters.com/> the property `education` could have a URL: https://criminalcharacters.com/vocab/#education which resolves to an HTML page that explains the term.
+
+For ad hoc terms where the crate author does not have the resources to put up an HTML page, an ad-hoc URL MAY be used in the form `https://w3id.org/ro/terms/criminalcharacters/education` where `criminalcharacters` is acting as a _namespace_ for one or more related terms like `education`. Ad-hoc namespaces under `https://w3id.org/ro/terms/` are available on first-come-first-serve basis; to avoid clashes, namespaces SHOULD be registered by [submitting terms and definitions](https://github.com/ResearchObject/ro-terms) to the RO-Crate project.
```
dgarijo, line 2027: Could the url be made to: https://w3id.org/ro/terms/criminalcharacters#education?

dgarijo: The reason is that doing content negotiation based on full files may be easier to handle than doing CN to individual terms.

ptsefton: I want to leave the example as it is a real one - there won't be any content negotiation but it is a reasonable strategy for someone to document their terms (and a machine readable version could be included in-page). 

Could you supplu an example that points to  https://w3id.org/ro/terms/ ?

PR #91, opened 2020-09-18
Remove HTML use of relatedLink property

stain:
The property `relatedLink` is not needed in the [RepositoryObject hasFile example](https://www.researchobject.org/ro-crate/1.0/#digital-library-and-repository-content), and the use of inline HTML for <http://schema.org/relatedLink> is not matching upstream spec, which expects `URL`.


stain:
In call 2020-09-24 @ptsefton agreed to merge - however the fact that this HTML "just worked" in Calcyte needs to be revisited as it can be a potential exploit factor.

PR #92, opened 2020-09-18
Docs fixes

simleo:
Fixes an id reference and some typos.

### DIFF
--- a/docs/0.2/index.md
+++ b/docs/0.2/index.md
@@ -195,7 +195,7 @@ Generally, the standard keys for [schema.org](http://schema.org) should be used.
 
 ### Additional metadata standards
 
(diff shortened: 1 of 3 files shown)


stain:
Thanks @simleo! Merged into the split versions from #95 

I did not change the past version as we would need to do a new patch release for typo fixes. We've tended to only fix syntactic bugs. 

I did not include 63f6c5fdf9af34b6726c7f7f89c032265dd5e1d9 as describing your own encoding format in a neighboring file sounds more like an edge case. It is true that the use of commonwl URI as example there is odd, as it will be an impedance mismatch with the use of `programmingLanguage` later in Workflows sections. Perhaps have a new issue for that.

PR #94, opened 2020-09-23
Attempt to distinguish contextual/data entities

stain:
.. also to reference forward about their identifiers, and what happens if a contextual entity is also a data entity.

### DIFF
--- a/docs/1.1-DRAFT/index.md
+++ b/docs/1.1-DRAFT/index.md
@@ -717,6 +717,24 @@ In all cases, consumers should be aware that a `DataDownload` is a snapshot that
 
 The _RO-Crate JSON-LD_ @graph SHOULD contain additional information about _Contextual Entities_ for the use of both humans (in `ro-crate-preview.html`) and machines (in `ro-crate-metadata.json`). This also helps to maximise the extent to which an _RO-Crate_ is self-contained and self-describing, in that it reduces the need for the consumer of an RO-Crate to refer to external information which may change or become unavailable over time.
 
+RO-Crate distinguishes between _Contextual entities_ and _Data entities_. 
+
+**Data entities** primarily exist in their own right as a file or directory (which may be in the _RO-Crate Root_ directory or downloadable by URL). 
+
+**Contextual entities** however primarily exist outside the digital sphere (e.g. [People](#people), [Places](#places)) or are  conceptual descriptions that primarily exists as metadata, like [GeoCoordinates] and [ContactPoint](#contact-information).
+
+A challenge can be how to assign [identifiers for contextual entities](#describing-entities-in-json-ld): 
+
+RO-Crate recommend that if an existing permalink (e.g. `https://orcid.org/0000-0002-1825-0097`) or other absolute URI (e.g. `https://en.wikipedia.org/wiki/Josiah_S._Carberry`) is reasonably unique for that entity, that URI should be used as identifier for the contextual entity in preference of an identifier local to the RO-Crate (e.g. `#josiah` or `#0fa587c6-4580-4ece-a5df-69af3c5590e3`). 
+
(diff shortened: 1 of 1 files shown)


stain:
Pre-emptily added in #95 - see https://www.researchobject.org/ro-crate/1.1-DRAFT/contextual-entities.html#contextual-vs-data-entities

@ptsefton or @nsjuty - are you able to review?

docs/1.1-DRAFT/index.md
```diff
@@ -717,6 +717,24 @@ In all cases, consumers should be aware that a `DataDownload` is a snapshot that
...
+
+Where a related URL exist, but is not unique enough to serve as identifier, it can instead be added to a locally identified contextual entity using [url].
+
+Care should be taken to not describe two conceptually different contextual entities with the same identifier - e.g. if `https://en.wikipedia.org/wiki/Josiah_S._Carberry` is a [Person] it should not also be a [CreativeWork] (although this example is a fictional person!).
+
+Some contextual entities can also be considered data entities - for instance the [license](#licensing-access-control-and-copyright) property refers to a [CreativeWork] that can reasonably be downloaded, however a license document is not usually considered as part of research outputs and would therefore typically not be included in [hasPart] on the _root data entity_. 
+
+Likewise, some data entities may also be described as contextual entities, for instance a `File` that is also a [ScholarlyArticle]. In such cases the _Contextual Data Entity_ SHOULD be described a single JSON object in the `@graph` with both types listed in a `@type` array. Consumers of an RO-Crate encountering two entities with the same `@id` SHOULD interpret them as a single entity.
```
simleo, line 736: How should multiple entities with the same `@id` be merged? For keys that allow multiple values such as `@type`, one could perform a set union of individual types, e.g.:

```json
{
  "@id": "https://doi.org/foo",
  "@type": "ScholarlyArticle"
}
```

plus

```json
{
  "@id": "https://doi.org/foo",
  "@type": "File"
}
```

equals

```json
{
  "@id": "https://doi.org/foo",
  "@type": ["File", "ScholarlyArticle"]
}
```

But what about properties where it doesn't make sense to have multiple values (e.g., `datePublished`)? should one override the others? I think this kind of flexibility puts a considerable burden on implementations (though I understand it's a SHOULD and not a MUST). A natural first step to process the metadata is to build a map to make entities retrievable by their `@id` (indeed, that's even recommended in the current implementation notes), but this becomes quite cumbersome if entities with the same `@id` are allowed and have to be merged. I think a better convention would be to state that, if multiple entities with the same `@id` do appear in the metadata file, the last one takes precedence. In this case, building the look-up table is trivial and efficient. For instance, it's a one-liner in Python:

```python
table = {e["@id"]: e for e in json_data["@graph"]}
```

Or perhaps just state that multiple entities with the same `@id` are not supported and if they are encountered the behavior will be undefined / implementation dependent.

ptsefton: I think trying merge is too complicated. Agree with the suggestion "Or perhaps just state that multiple entities with the same @id are not supported and if they are encountered the behavior will be undefined / implementation dependent."

ptsefton (changes requested):
This is largely OK, but I think this bit is problematic:

" Consumers of an RO-Crate encountering two entities with the same `@id` SHOULD interpret them as a single entity."

See my reply to @simleo below - we should strongly recommend NOT duplicating IDs and say that if you do then the behaviour is undefined.

docs/1.1-DRAFT/index.md
stain: Agree on saying it is undefined. Someone doing it as JSON-LD could try that, but it would get messy in JSON; so let's just say it is undefined, and neither would be at fault if they merged or dropped the nodes.

stain: So if they are not supported, we would need to change `SHOULD` to `MUST`?

Issue #65, opened 2019-11-20
Change CSS/theme to wider view or ReadTheDocs style


stain:
**Describe the bug**

The RO Crate 1.0 specification page has quite a narrow page due to the GitHub Pages style. While it looks nice it can be a bit of a long scroll, and the narrow margin is not so well suited for some of the JSON-LD examples.

Ironically if viewed on a small device in portrait mode, the empty white space left column disappears and the then wider text is much easier to read.

**URL**
https://w3id.org/ro/crate/1.0/

**Suggested fix**

We could use a simpler Bootstrap theme, like on <https://rawcdn.githack.com/ResearchObject/ro-crate/4332b100dd651c3b1cde840eaf8b4b86ad999bca/docs/1.0/ro-crate-1.0.0.html>

..but then there won't be logo and link back to the front page etc. 

A floating table-of-content that updates as you scroll would be nice thing to have on the side.

I am not sure if I am suggesting to put a hard-coded HTML like the above, or to change the Jekyll style for the whole RO-Crate site (which would affect all pages).

alaninmcr:
At the moment, it makes poor use of "real estate". I'd recommend starting with the simpler Bootstrap and then add in any necessary links.

marcolarosa:
+1

stain:
@ljgarcia suggested using a ReadTheDocs style theme, see for instance https://isa-specs.readthedocs.io/en/latest/isajson.html

That would look very nice! And we could do a similar page split of the main sections, e.g. Contextual Entities, each of the appendices, etc.

I am not sure how we can activate that in GitHub pages as a theme, at least I hope we don't need to rewrite the Markdown to RST! Do @ljgarcia have any suggestions?

I also think if we change the markdown theme it would change the look on the old https://researchobject.github.io/ro-crate/1.0/ - however if this breaks we could save/archive 1.0 and older as the existing HTML rather than MarkDown.

PR #95, opened 2020-09-23
Use jekyll-rtd-theme and split 1.1-DRAFT into files per section

stain:
As suggested in #65 we need a better theme that shows the examples well/wide, and that has good navigation between sections.

This pull request changes the RO-Crate website to use the [jekyll-rtd-theme](https://github.com/rundocs/jekyll-rtd-theme) by <https://rundocs.io/>

It has the same look-and-file as <https://readthedocs.org/> with a table of content on the left - but working with Jekyll/Markdown instead of transitioning to Python/Sphinx/RST. It seems like quite an improvement for RO-Crate in terms of readability and navigation:

rtd

You can try it on https://stain.github.io/ro-crate/1.1-DRAFT/


stain:
To improve navigation I split out into multiple files with section numbering in TOC. For instance [root data entity](https://stain.github.io/ro-crate/1.1-DRAFT/root-data-entity.html) is a separate page.

Some caveats with this theme:

1. `index.md` / `README.md` in a folder is used mainly to add top-level categories and only navigable from the breadcrumbs - I therefore added an [about](https://stain.github.io/ro-crate/1.1-DRAFT/about.html) page which is kind of repeated as _abstract_ on the top-level [`1.1/` root](https://stain.github.io/ro-crate/1.1-DRAFT/) but including a list of subpages.
2. All folders are shown in sidebar, including old drafts. This would get messy, so I added an `exclude: true` patch
3. One of the jekyll plugins is adding github links like @stain to text starting with `@` - obviously `@graph` is not meaning a github user! Work-around - using `  `` ` backtick quotes
4. I seem to have to copy that bit `[RFC1234]: http://example.com/RFC1234` URL list at bottom into every Markdown file. Obviously each of the pages only refer to a couple of those links but I've not trimmed them down.

I had to rearrange some of the section headings as before it was all under a massive _Contextual entities_. I tried to update all internal links that now go across pages now, but I may have got some wrong. Check if it makes sense.

I only restructured the new `1.1-DRAFT/` (to be `1.1/`) so the question is what to do with the older single-page [0.2](https://stain.github.io/ro-crate/0.2/) and [1.0](https://stain.github.io/ro-crate/1.0/) - it seems they render kind-of OK, and I found a way to exclude them from the side-bar, so I suggest leaving them as-is.  An alternative would be to is to save them as HTML with the old theme from <https://stain.github.io/ro-crate/1.0/> - One important thing is to ensure the `1.0` permalinks and internal links still work - therefore I would not want to split up these as the section `#headings` would then move.

In this style we could also move some of the frontpage homepage stuff to subpages, e.g. "Implementations" and "Community". I would also add "Specifications" to list the older versions there.

Now we should not really merge this PR until we are agreed to release 1.1 as this would "promote" the "1.1-DRAFT" to the top-level menu.  I would use that `exclude` mechanism again to hide all except the current version.

stain:
In #65 @ljgarcia suggested trying [just-the-docs](https://pmarsceill.github.io/just-the-docs/) which looks really nice and modern in comparison:

justthedocs

However I found this a bit cumbersome. I have to annotate `parent` in each of the "child" pages - the navigation hierarchy does not have to match file system. It is more flexible, e.g. the exclude thing is built in. 

Biggest drawback for me was that the internal section headings don't appear on the left and has to be added with a classical `{:toc}` as in the screenshot. This is a great feature of readthedocs-style themes that we need, at least for the [contextual entities_ pages](https://stain.github.io/ro-crate/1.1-DRAFT/contextual-entities.html) which we would otherwise have to split into even smaller pages, but that would ruin some of the flow (e.g. Organization is after Person)

ljgarcia:
It looks nice, indeed an improvement. I tried the [testing version](https://stain.github.io/ro-crate/1.1-DRAFT/data-entities.html) provided by @stain and all links I tried worked for me. I like the previous and next buttons at the end of the page. What I am not so sure is the pencil icon on the top right. Do you want to encourage editions (via fork/PR) that directly? I do not expect much people will use this option, not sure if needed.

There is also a sort of blink when navigating with the menu. When changing to another menu item, the page is re-loaded and I can see like a shadow of the logo. I guess it goes to the top of the menu but then immediately moves to where the menu was clicked. It is not a real issue so if it comes from the template I would leave it but maybe worth to have a quick look at it.

I tried it out in Windows 10, Firefox and Chrome.

ptsefton:
Agree with @ljgarcia - looks good. IS there a way to get it as a single file without the navigation chopping it up into pages? Not sure if this is an issue?

stain:
Discussed in [call 2020-09-24](https://s.apache.org/ro-crate-minutes#heading=h.u1i17jpqpnat) where it was agreed with @ptsefton @dgarijo @CaroleGoble @stain and the silent majority to change the theme.

1.1 is however in review for another week (see #97), so I'll merge this so reviewers can edit the splitted files; after setting the "exclude" flags so `1.1-DRAFT` does not dominate and merging #93

Issue #97, opened 2020-09-24
Review RO-Crate 1.1-DRAFT new sections


stain:
Following discussion in call [2020-09-24](https://s.apache.org/ro-crate-minutes#heading=h.wf9ll2ymr0qu) we will review [RO-Crate 1.1-DRAFT](https://www.researchobject.org/ro-crate/1.1-DRAFT/#table-of-contents) until Thurs 2020-10-01, as we have added new sections that have not been proofread yet.

After review we aim to release 1.1 on Fri 2020-10-02.

From [Changelog](https://www.researchobject.org/ro-crate/1.1-DRAFT/appendix/changelog.html):

  * **Note**: The RO-Crate metadata file is renamed to `ro-crate-metadata.json` to facilitate use of JSON editors.  [#82](https://github.com/ResearchObject/ro-crate/issues/82) [#84](https://github.com/ResearchObject/ro-crate/issues/84)
  * [Data entities](https://github.com/ResearchObject/ro-crate/tree/master/docs/1.1-DRAFT/data-entities.md) can reference external resources with absolute URI [#74](https://github.com/ResearchObject/ro-crate/issues/74)
  * Added section on considerations for [Web-based Data Entities](https://github.com/ResearchObject/ro-crate/tree/master/docs/1.1-DRAFT/data-entities.md#web-based-data-entities)  [#74](https://github.com/ResearchObject/ro-crate/issues/74)
  * The [root dataset](https://github.com/ResearchObject/ro-crate/tree/master/docs/1.1-DRAFT/root-data-entity.md#direct-properties-of-the-root-data-entity) is no longer required to be `./` [#74](https://github.com/ResearchObject/ro-crate/issues/74)
  * [RO-Crate Root](https://github.com/ResearchObject/ro-crate/tree/master/docs/1.1-DRAFT/structure.md) directory no longer requires payload files [#74](https://github.com/ResearchObject/ro-crate/issues/74)
  * [Workflows and scripts](https://github.com/ResearchObject/ro-crate/tree/master/docs/1.1-DRAFT/workflows.md) section now aligned with [BioSchemas ComputationalWorkflow profile](https://bioschemas.org/profiles/ComputationalWorkflow/0.5-DRAFT-2020_07_21/)  [#81](https://github.com/ResearchObject/ro-crate/issues/81)
[...]

PR #100, opened 2020-09-29
Reflect the MUST/SHOULD from BioSchemas profile

stain:
Add all MUST/SHOULD/COULD from https://bioschemas.org/profiles/ComputationalWorkflow/0.5-DRAFT-2020_07_21/ and https://bioschemas.org/profiles/FormalParameter/0.1-DRAFT-2020_07_21/

This does get quite verbose and I had to take some liberties on the examples as BioSchemas did not provide those.

What I think is important to highlight is that it is not necessary for RO-Crate users doing workflows to comply with the BioSchemas profile - in fact by listing all of these we need perhaps to downgrade from SHOULD to OPTIONAL?

See issue #97 and comment https://github.com/ResearchObject/ro-crate/issues/97#issuecomment-700117827 by @alaninmcr


alaninmcr (changes requested):
The description for FormalParameter is split up strangely and sort-of duplicates. The descriptions must be made contiguous

stain:
@alaninmcr In 40dcf3050d3fd1823ed9ece877c78730c79734ce I tried changing the language, but I'm not sure how to avoid splitting it up as they are different MUST/SHOULD/MAY, and the above `ComputationalWorkflow` has a similar split.

docs/1.1-DRAFT/workflows.md
```diff
@@ -169,26 +175,70 @@ A workflow diagram may still be provided even if there is no programmatic `Softw
...
-* [sdPublisher] to identify the [Person] or [Organization] who has made the JSON-LD description of the workflow
-* [url] to identify a public webpage or source code repository for the workflow
+* [sdPublisher] to identify the [Person] or [Organization] who has made this JSON-LD description of the workflow
+* [url] to identify a public page or source code repository for the workflow
 * [version] to indicate the released version of this workflow file
 
-Contextual entities for [FormalParameter], referenced by `input` or `output`, SHOULD describe:
+The [ComputationalWorkflow] data entity SHOULD also describe, where relevant:
+* [citation] to identify a [CreativeWork] that should be cited for this workflow
+* [contributor] to identify [Person] or [Organization] that helped conceive the workflow
```
dgarijo, line 196: What about the author? Contributor may have a little loose meaning and people often distinguish both

alaninmcr (approved):
Looks good to me. Go for it

docs/1.1-DRAFT/workflows.md
dgarijo: I see that author is already included. Resolving conversation

docs/1.1-DRAFT/workflows.md
```diff
@@ -169,26 +175,70 @@ A workflow diagram may still be provided even if there is no programmatic `Softw
...
+* [targetProduct] to indicate a [SoftwareApplication] for the operating system or product which this code applies to, e.g. `https://www.microsoft.com/en-gb/windows/get-windows-10` or `https://releases.ubuntu.com/20.04/`
+
+The [ComputationalWorkflow] data entity MAY also describe, where relevant:
+* [alternateName] to indicate an alternative [name] that help identify the workflow, e.g. `nf-core/chipseq`
+* [conditionsOfAccess] to indicate restrictions on the availability of this workflow, e.g. `"To protect endangered wildlife, this biodiversity workflow is only available to approved research projects"`
+* [dateModified] to indicate a date later than [dateCreated] when this workflow was modified
+* [datePublished] to indicate a date different from [dateCreated] for when this workflow was made public
+* [encodingFormat] to indicate the [IANA media type](data-entities.md#adding-detailed-descriptions-of-encodings), e.g. `"application/xml"` or `"application/json"` (see also [programmingLanguage] above)
+* [image] to indicate an [ImageObject] that represents the workflow, e.g. a workflow diagram
```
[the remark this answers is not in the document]
dgarijo: I see an example in the spec above, but maybe having an example to an online existing image would be helpful too

docs/1.1-DRAFT/workflows.md
```diff
@@ -169,26 +175,70 @@ A workflow diagram may still be provided even if there is no programmatic `Softw
...
+* [identifier] to provide a unique identifier for this workflow, e.g. `https://doi.org/10.5281/zenodo.3966161` or `urn:uuid:0a98135b-3e48-4f76-aed8-4a7f6421bd74`
+
+### Describing inputs and outputs
+
+The input and output _parameters_ for a workflow or script can be given with `input` and `output` to [FormalParameter]
```
dgarijo, line 221: I would be careful with the word "parameter" here. Some people may understand parameters as variables (numerical values, or strings). However above `inputs` and `outputs` are used to refer to files. Which one is it? Is it both?

dgarijo: Also, you cannot have a numerical parameter as an output, unless it's in a file.

docs/1.1-DRAFT/workflows.md
```diff
@@ -169,26 +175,70 @@ A workflow diagram may still be provided even if there is no programmatic `Softw
...
+contextual entities. Note that this entity usually represent a _potential_ input/output value in a reusable
+workflow, much like [function parameter definitions](https://en.wikipedia.org/wiki/Parameter_(computer_programming))
+in general programming.
+
+When complying with the
+[BioSchemas FormalParameter profile](https://bioschemas.org/profiles/FormalParameter/0.1-DRAFT-2020_06_23/)
+the _contextual entities_ for [FormalParameter], referenced by `input` or `output`, MUST describe:
 
 * [name] given the programmatic name for the parameter binding
```
dgarijo, line 230: First time the notion of parameter binding is included, but it's not defined. Maybe an example would clarify things here

docs/1.1-DRAFT/workflows.md
```diff
@@ -169,26 +175,70 @@ A workflow diagram may still be provided even if there is no programmatic `Softw
...
 * [additionalType] identifying the most specific subtype of [EDAM Data](http://edamontology.org/data_0006) (fallbacks [Data](http://edamontology.org/data_0006) or [Text data](http://edamontology.org/data_2526))
 * [encodingFormat] identifying the most specific subtype of [EDAM Format](http://edamontology.org/format_1915) (fallbacks [Binary format](http://edamontology.org/format_2333) or [Textual format](http://edamontology.org/format_2330))
+
```
dgarijo, line 233: EDAM was designed in the bioinformatics space. What if the endocoding formats or types there are not enough? Can I suggest my own data type?

docs/1.1-DRAFT/workflows.md
```diff
@@ -169,26 +175,70 @@ A workflow diagram may still be provided even if there is no programmatic `Softw
...
+
+In addition the `FormalParameter` contextual entities SHOULD also include:
+* [description] describing the parameter's expected value range and purpose
+
+Contextual entities of type `FormalParameter` MAY also include any of:
 * [valueRequired] `true` if this (input) parameter must be specified to run the workflow, or `false` (default) if parameter is optional.
 * [defaultValue] present if this (input) parameter has a default value. In RO-Crate this SHOULD be in the form of a `"string"` or a `{"@id": "data/entity.txt"}`
+* [identifier] if the parameter has a global identifier that persists across workflows it is used in.
```
dgarijo, line 241: I do not understand very well this field. Why is it needed?

docs/1.1-DRAFT/workflows.md
```diff
@@ -169,26 +175,70 @@ A workflow diagram may still be provided even if there is no programmatic `Softw
...
 * [additionalType] identifying the most specific subtype of [EDAM Data](http://edamontology.org/data_0006) (fallbacks [Data](http://edamontology.org/data_0006) or [Text data](http://edamontology.org/data_2526))
 * [encodingFormat] identifying the most specific subtype of [EDAM Format](http://edamontology.org/format_1915) (fallbacks [Binary format](http://edamontology.org/format_2333) or [Textual format](http://edamontology.org/format_2330))
+
+
+In addition the `FormalParameter` contextual entities SHOULD also include:
+* [description] describing the parameter's expected value range and purpose
+
+Contextual entities of type `FormalParameter` MAY also include any of:
 * [valueRequired] `true` if this (input) parameter must be specified to run the workflow, or `false` (default) if parameter is optional.
 * [defaultValue] present if this (input) parameter has a default value. In RO-Crate this SHOULD be in the form of a `"string"` or a `{"@id": "data/entity.txt"}`
```
dgarijo, line 240: Why not add minimum and maximum accepted values? (Maybe in a future iteration?)

dgarijo (changes requested):

docs/1.1-DRAFT/workflows.md
```diff
@@ -169,26 +175,70 @@ A workflow diagram may still be provided even if there is no programmatic `Softw
...
-* [url] to identify a public webpage or source code repository for the workflow
+* [sdPublisher] to identify the [Person] or [Organization] who has made this JSON-LD description of the workflow
+* [url] to identify a public page or source code repository for the workflow
 * [version] to indicate the released version of this workflow file
 
-Contextual entities for [FormalParameter], referenced by `input` or `output`, SHOULD describe:
+The [ComputationalWorkflow] data entity SHOULD also describe, where relevant:
+* [citation] to identify a [CreativeWork] that should be cited for this workflow
+* [contributor] to identify [Person] or [Organization] that helped conceive the workflow
+* [creativeWorkStatus] to indicate the development state of the workflow, e.g. `"Draft"` or `"Published"`
```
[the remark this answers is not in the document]
stain: Yes, I did not extend here from what [Bioschemas lists for `creativeWorkStatus`](https://bioschemas.org/profiles/ComputationalWorkflow/0.5-DRAFT-2020_07_21/) - but those example names are more relevant for textual creative works and perhaps Bioschemas should be more specific. 

It is free-text though (or controlled vocabulary URIs) so I'm OK with adding our our own `"Released"` - "latest version" is a bit pointless as every version at the point of their release would claim to be the latest version.

docs/1.1-DRAFT/workflows.md
```diff
@@ -169,26 +175,70 @@ A workflow diagram may still be provided even if there is no programmatic `Softw
...
 * [name] giving a short descriptive human-readable name for the workflow
 * [programmingLanguage] identifying the workflow system, typed as `ProgrammingLanguage`
 * [creator] identifying the [Person](s) and/or [Organization](s) that assembled the workflow. (This may differ from the [author] of the conceptual workflow, e.g. if the workflow was converted from one language to another).
 * [dateCreated] the date the workflow was first created, e.g. `2020-05-23`
 * [license] identifying a [CreativeWork] that details the license for distributing or editing the workflow
 * [input] and [output] identifying contextual entities for the [FormalParameter] describing input and output parameters/variables that may be varied on different workflow executions
-* [sdPublisher] to identify the [Person] or [Organization] who has made the JSON-LD description of the workflow
-* [url] to identify a public webpage or source code repository for the workflow
+* [sdPublisher] to identify the [Person] or [Organization] who has made this JSON-LD description of the workflow
+* [url] to identify a public page or source code repository for the workflow
```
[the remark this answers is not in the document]
stain: Schema.org has this awkward split between https://schema.org/SoftwareApplication and https://schema.org/SoftwareSourceCode but many useful properties are only on `SoftwareApplication` or even `WebAPI`.

Perhaps Bioschemas can propose that https://schema.org/documentation also applies to `SoftwareApplication` and `SoftwareSourceCode` rather than just https://schema.org/WebAPI - it would also make sense for `documentation` to be subproperty of https://schema.org/subjectOf  (it might also document other things)

There are many potential http://schema.org/CreativeWork fields like:

* https://schema.org/url - the homepage of the workflow itself <https://nf-co.re/chipseq>
* https://schema.org/identifier - a persistent identifier for workflow e.g. DOI <https://zenodo.org/record/3966161>
* https://schema.org/citation - a scholarly article that (ideally) talks about this workflow e.g. <https://doi.org/10.1038/s41587-020-0439-x.>
* https://schema.org/mainEntityOfPage (referencing a page that is mainly about this workflow - e.g. documentation <https://biobb.readthedocs.io/> or registration <https://workflowhub.eu/workflows/56>
* https://schema.org/discussionUrl (a place to discuss this workflow - e.g. <https://ask.bioexcel.eu/c/BioExcel-Building-Blocks-library/23> or even GitHub
* https://schema.org/subjectOf - documentation page that talks about this workflow, but also other things <http://mmb.irbbarcelona.org/biobb/workflows>
* https://schema.org/isBasedOn - another workflow/pipeline it is based on, e.g. <http://www.mdtutorials.com/gmx/lysozyme/index.html>
* https://schema.org/usageInfo - tutorials e.g. <http://mmb.irbbarcelona.org/biobb/workflows/tutorials/md_setup>
* https://schema.org/workExample - more specific example that realize this workflow e.g. <https://github.com/bioexcel/pmxlaunchCV19/blob/master/workflows/CV19/pmx_cv.py> realizes the tutorial workflow by adding specific data

And from SoftwareSourceCode:
* https://schema.org/codeRepository - GitHub repo (root?) e.g. <https://github.com/nf-core/chipseq>

Example of needing several of these fields in WorkflowHub for the BioBB workflow, under _Discussion channels_ on the right of https://workflowhub.eu/workflows/56:

discussion-channels

@dgarijo if we want to add how to link to documentation, we would need to add it in the section above bioschemas conformance, as [Bioschemas ComputationalWorkflow](https://bioschemas.org/profiles/ComputationalWorkflow/0.5-DRAFT-2020_07_21/) does not have a view on this yet. Do you think we add this now, or we wait for next release of ComputationalWorkflow profile?

@alaninmcr  shall we feed this back into the next BioSchemas profile on how to link to workflow documentation? Make issue(s) in https://github.com/BioSchemas/specifications

docs/1.1-DRAFT/workflows.md
```diff
@@ -169,26 +175,70 @@ A workflow diagram may still be provided even if there is no programmatic `Softw
...
+* [keywords] with comma separated list of keywords that categorise this workflow
+* [maintainer] to indicate a [Person] or [Organization] responsible for maintaining the workflow, specially if different from [creator]
+* [producer] the [Organization] or [Person] that, when different from [creator] and [maintainer], can be considered to _produce_ the workflow; typically a [Project]
+* [publisher] the [Organization] or [Person] that is making the workflow publicly available, e.g. `https://workflowhub.eu/`, `https://usegalaxy.eu/` or `https://usegalaxy.eu/`
+* [runtimePlatform] the runtime platform or script interpreter as an informal text string, e.g. `"Python2.3"`, `"Galaxy"` (see also [programmingLanguage] above)
+* [softwareRequirements] indicating [SoftwareApplication]s or [SoftwareSourceCode]s needed for this workflow, its steps and workflow engine to execute, e.g. `https://bioconda.github.io/`, `https://docs.docker.com/engine/`, `https://bio.tools/samtools`
+* [targetProduct] to indicate a [SoftwareApplication] for the operating system or product which this code applies to, e.g. `https://www.microsoft.com/en-gb/windows/get-windows-10` or `https://releases.ubuntu.com/20.04/`
```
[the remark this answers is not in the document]
stain: Yes, we could perhaps think of https://schema.org/runtimePlatform as the textual version of https://schema.org/targetProduct - it is much more important to formally link to the runtime for the workflow rather than the operating system that engine again runs on - but then many workflows will also be platform-specific.  

For instance you can run Snakemake on Windows - but a particular snakemake workflow, which runs shell commands, will probably require OSX/Linux or even a particular distribution if it is not using conda packages for dependencies. 

We are here limited to listing properties from [ComputationalWorkflow](https://bioschemas.org/profiles/ComputationalWorkflow/) profile but I added the examples -  agree we can make more workflow-specific examples.


So perhaps we need both and the workflow engine itself should be under `targetProduct` (the application that can run this workflow) while `softwareRequirements` can list what the workflow and engine needs to run it (e.g. nextflow as product would need the conda requirement for this particular workflow)

docs/1.1-DRAFT/workflows.md
```diff
@@ -169,26 +175,70 @@ A workflow diagram may still be provided even if there is no programmatic `Softw
...
+* [isBasedOn] to indicate another [ComputationalWorkflow] or [CreativeWork] this workflow is based on or derived from
+* [keywords] with comma separated list of keywords that categorise this workflow
+* [maintainer] to indicate a [Person] or [Organization] responsible for maintaining the workflow, specially if different from [creator]
+* [producer] the [Organization] or [Person] that, when different from [creator] and [maintainer], can be considered to _produce_ the workflow; typically a [Project]
+* [publisher] the [Organization] or [Person] that is making the workflow publicly available, e.g. `https://workflowhub.eu/`, `https://usegalaxy.eu/` or `https://usegalaxy.eu/`
+* [runtimePlatform] the runtime platform or script interpreter as an informal text string, e.g. `"Python2.3"`, `"Galaxy"` (see also [programmingLanguage] above)
+* [softwareRequirements] indicating [SoftwareApplication]s or [SoftwareSourceCode]s needed for this workflow, its steps and workflow engine to execute, e.g. `https://bioconda.github.io/`, `https://docs.docker.com/engine/`, `https://bio.tools/samtools`
```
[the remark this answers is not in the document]
stain: Agree - let's raise processor and memory requirements in <https://github.com/BioSchemas/specifications/issues>

docs/1.1-DRAFT/workflows.md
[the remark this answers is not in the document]
stain: It's confusing throughout, because you also have `@id`!  

I think the nf-core/chipseq example would separate it good:

```json
{
  "@id": "https://raw.githubusercontent.com/nf-core/chipseq/1.2.1/main.nf",
  "@type": ["CreativeWork", "SoftwareSourceCode", "ComputationalWorkflow"],
  "name": "ChIP-seq peak-calling, QC and differential analysis pipeline",
  "alternateName": "nf-core/chipseq",
  "url": "https://nf-co.re/chipseq/1.2.1",
  "version": "1.2.1",
  "codeRepository": {"@id": "https://github.com/nf-core/chipseq"},
  "identifier": "https://doi.org/10.5281/zenodo.3966161",
  "citation": {"@id": "https://doi.org/10.1038/s41587-020-0439-x"}
}
```

For https://schema.org/identifier and https://schema.org/identifier throughout the RO-Crate spec I think we have made them all strings rather than `@id`

docs/1.1-DRAFT/workflows.md
alaninmcr: Yes - it could be a URL to elsewhere

docs/1.1-DRAFT/workflows.md
alaninmcr: They would only realistically apply to numeric parameters. It would need more thought - and yes a change to bioschemas

docs/1.1-DRAFT/workflows.md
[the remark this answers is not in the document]
alaninmcr: Yes a lot of these seem to be suggestions for the ComputationalWorkflow profile, rather than about RO-Crate. Is there a better way of separating the two?

docs/1.1-DRAFT/workflows.md
```diff
@@ -169,26 +175,70 @@ A workflow diagram may still be provided even if there is no programmatic `Softw
...
-Contextual entities for [FormalParameter], referenced by `input` or `output`, SHOULD describe:
+The [ComputationalWorkflow] data entity SHOULD also describe, where relevant:
+* [citation] to identify a [CreativeWork] that should be cited for this workflow
+* [contributor] to identify [Person] or [Organization] that helped conceive the workflow
+* [creativeWorkStatus] to indicate the development state of the workflow, e.g. `"Draft"` or `"Published"`
+* [description] to further describe the workflow's purpose and method
+* [funding] to a [Grant] contextual entity that has financially supported development of the workflow
+* [hasPart] to indicate individual steps as [SoftwareApplication], [SoftwareSourceCode] or [ComputationalWorkflow] entities
+* [isBasedOn] to indicate another [ComputationalWorkflow] or [CreativeWork] this workflow is based on or derived from
+* [keywords] with comma separated list of keywords that categorise this workflow
```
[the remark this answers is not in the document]
alaninmcr: As you may know, it is a single field that is comma separated. I don't like it, but it is inherited from schema.org

stain:
Discussing this with @alaninmcr I think this PR is adding a bit too much by duplicating the Bioschemas profile into RO-Crate - which can cause confusion on authority. Lots of the nice feedback from @dgarijo apply to the profile rather than RO-Crate - so perhaps we should soften our language and redudc this section down to just referencing the profile, list it's MUST requirements inline, and give the example.

docs/1.1-DRAFT/workflows.md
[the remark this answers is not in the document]
dgarijo: The citation can be text. In order to make it simpler, why not add the citation as "Ewels, P.A., Peltzer, A., Fillinger, S. et al. The nf-core framework for community-curated bioinformatics pipelines. Nat Biotechnol 38, 276–278 (2020). https://doi.org/10.1038/s41587-020-0439-x"

docs/1.1-DRAFT/workflows.md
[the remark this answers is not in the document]
dgarijo: I don't understand very well. Why there? We can just add them here, right? It looks like a bottleneck to wait for others to include it in bioschemas when they are already in schema.org.

docs/1.1-DRAFT/workflows.md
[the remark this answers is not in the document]
dgarijo: +1. The line becomes a little fuzzy. I understand that when you are talking on code/software, it makes sense to have the relationship through targetProduct. E.g., this code is for the software "BioToolsYZ1.0", but when it comes to workflows, maybe it's just too much.

docs/1.1-DRAFT/workflows.md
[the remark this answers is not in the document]
dgarijo: @alaninmcr I know this. You know this. But the reader may not know this :) I think it is best for adoption if we don't assume the reader is a schema.org expert

docs/1.1-DRAFT/workflows.md
dgarijo: Awesome

docs/1.1-DRAFT/workflows.md
dgarijo: In our work, having the following properties for parameters has been incredibly helpful:
- minimum value (min numeric value or date)
- maximum value (max numeric value or date)
- accepted values (i.e., this parameter is a string that accepts "Maize" and "Sorghum" as values)
- interval (i.e., create an ensemble increasing this parameter by 0.1 until the max value)

dgarijo:
@stain happy to review after the section has been reduced. I just want to clarify that these changes are all very good. My intention was not to block them, just to clarify them for potential adopters.

docs/1.1-DRAFT/workflows.md
alaninmcr: It is clear in the bioschemas documentation. I think this can be closed.

docs/1.1-DRAFT/workflows.md
stain: Changed to just "parameter"

docs/1.1-DRAFT/workflows.md
stain: Removed our expanded list, reduced to just referencing their MUST and SHOULD fields:

> When complying with this profile, the workflow data entities MUST describe these properties and their related contextual entities:  [name], [programmingLanguage], [creator], [dateCreated], [license], [sdPublisher], [url], [version].

> The [ComputationalWorkflow profile] specifies additional properties that a compliant [ComputationalWorkflow] data entity SHOULD include: [citation], [contributor], [creativeWorkStatus], [description], [funding], [hasPart], [isBasedOn], [keywords], [maintainer], [producer], [publisher], [runtimePlatform], [softwareRequirements], [targetProduct]

stain:
In 9f8ee23 I reduced this BioSchemas text significantly down - see rendering in https://stain.github.io/ro-crate/1.1-DRAFT/workflows.html#complying-with-bioschemas-computational-workflow-profile

Instead of bullet points with the long explanation (which sometimes extended the upstream profile) I now just reference the profile. The properties will have to be (better) explained by the profile at BioSchemas.org and can be further discussed there.

For the input/outputs https://stain.github.io/ro-crate/1.1-DRAFT/workflows.html#describing-inputs-and-outputs again I removed EDAM references for `additionalType` and `encodingFormat` as they seem to imply you can't use `FormalParameter` for non-Bio stuff.

PR #103, opened 2020-09-30
Review of section ## Web-based Data Entities

ljgarcia:
* Original text, line 193 "RO-Crates can also gather resources on the web". Proposed text "RO-Crates can also gather resources from the web". But I am not sure what you mean here. Is it gather as in collect? Or is it include pointers to resources on the web?
* "traversal", line 288. Not sure traversal or transversal is intended here, in any case, should it not be before download as an adjective?


stain (approved):

stain:
Thank you, @ljgarcia! I think RO-Crate gathering from the web is good - obviously resources external will stay where they are, yet by adding them in `hasPart` we are promoting them from not just being something out there to also being included as a part of composing this RO for whatever reason. So it is in a sense something between composition and reference!

PR #105, opened 2020-09-30
Review of Draft 1.1

dgarijo:


ptsefton:
I have a question about the metadata file descriptor: http://www.researchobject.org/ro-crate/1.1-DRAFT/root-data-entity.html#ro-crate-metadata-file-descriptor

The algorithm for finding the root dataset could be simplified by finding an item with "@id" = "ro-crate-metadata.json" - given that that is always the @id - rather than matching on the value of @conformsTo.

ptsefton (approved):
Looks good

dgarijo:
@ptsefton,
sorry, I just saw your comment. Is that addressed to me? I don't think I have an answer, sorry.

stain:
The reason @ptsefton for not matching on `@id` is because after parsing as JSON-LD the relative URI may have become absolute. This could also depend on symbolic links, HTTP redirections etc. See https://www.researchobject.org/ro-crate/1.1-DRAFT/appendix/relative-uris.html#expandingparsing-json-ld-keeping-relative-referencing

It does seem to imply that there should not be any other entities with that particular `conformsTo` - e.g. if there is a nested RO-Crate, how should we point to its JSON-LD file for further descriptions?

Issue #109, opened 2020-10-22
VOTE: Release RO-Crate 1.1.0 RC1


stain:
I am pleased to be calling this vote for the release of
**RO-Crate specification 1.1** 

Votes are open until [Monday 2020-10-26 17:00 UTC](https://everytimezone.com/s/11ad82e9).

The release candidate to be voted over is available at:

* HTML: <https://stain.github.io/ro-crate/1.1-RC1/index.html> 
 (to become `https://w3id.org/ro/crate/1.1` redirecting to `https://www.researchobject.org/ro-crate/1.1/`)
* Single page HTML: <https://github.com/ResearchObject/ro-crate/releases/download/1.1.0-RC1/ro-crate-1.1.0-RC1.html>
* Single page PDF: <https://github.com/ResearchObject/ro-crate/releases/download/1.1.0-RC1/ro-crate-1.1.0-RC1.pdf>
* JSON-LD context: <https://stain.github.io/ro-crate/1.1-RC1/context.jsonld> 
 (to become `https://w3id.org/ro/crate/1.1/context` redirecting to `https://www.researchobject.org/ro-crate/1.1/context.jsonld`)
* RO-Crate: <https://stain.github.io/ro-crate/1.1-RC1/ro-crate-preview.html> <https://stain.github.io/ro-crate/1.1-RC1/ro-crate-metadata.json>
* Changelog and downloads: https://github.com/ResearchObject/ro-crate/releases/tag/1.1.0-RC1

Please vote and/or comment on any bugs on this github issue #109 **before Monday** 2020-10-26 17:00 UTC. (_You can use "GitHub reaction" to vote_)

Anyone can participate in testing and voting, not just [RO-Crate community](https://www.researchobject.org/ro-crate/community.html) members, please feel free to check the release candidate and provide your votes.

tag: https://github.com/ResearchObject/ro-crate/releases/tag/1.1.0-RC1
commit: c9aa82ad2a7689907f16aaa0ea4f7e517a065a9a

stain:
I'm also preparing the Zenodo DOI with the release PDF/HTML. 

**Author list**: This is simply the https://www.researchobject.org/ro-crate/community members at time of release.

Note: @simleo just joined the team, but I had not updated the file in time for this RC1. Will include in final author list.

We should probably also acknowledge contributions in discussions from @marcolarosa although he did not sign up on #1 - anyone else?

dgarijo:
@stain,
[...]

PR #113, opened 2020-10-26
More about BagIt and fetch.txt

stain:
Adressing @jmfernandez in https://github.com/ResearchObject/ro-crate/issues/109#issuecomment-716204520
to also explain how BagIt `fetch.txt` relates to RO-Crate `uri`


jmfernandez (approved):
Very nice example @stain !

PR #127, opened 2021-02-03
Bug intro issue 125

### LINKED ISSUE #125: Introduction wrongly emphasizes file-based data; other types of data entity are also valid
**Describe the bug**
"RO-Crate method: “organizing **file-based** data with associated metadata, using linked data principles, in both human and machine readable formats, with the ability to include additional domain-specific metadata”" 

Thise  emphasises the file approach not the references. Without references it cannot support FDO and many other use cases.
That quote means that anyone not using files and not using data will tink RO-Crate is inappropriate. I can point to an entry in a database - that is not file-based data - it in Github. Linked Data principles is too obtuse - references needs to be explicit.

Its damaging, confusing and potentially will stop adoption and funding. Its also contrary to our presentations. 

**URL**
https://www.researchobject.org/ro-crate/1.1/introduction.html

**Suggested fix**
update the text

**Additional context**

ptsefton:
Proposed patch to correct draft  - fixes #125

### DIFF
--- a/docs/1.1/introduction.md
+++ b/docs/1.1/introduction.md
@@ -1,6 +1,8 @@
 ---
 layout: default
 title: Introduction
+redirect_from:
+  - /1.1-DRAFT/introduction
 sort: 2
 excerpt: |
   ...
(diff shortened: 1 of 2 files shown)


stain:
Somehow your b2eb3ff reverted the initial change, so I re-applied it to both 1.1 and 1.2-DRAFT.

stain (approved):

PR #134, opened 2021-03-09
New theme: just the docs

### LINKED ISSUE #123: Exclude *-DRAFT pages from search
Pages on https://www.researchobject.org/ro-crate/ do not show the current `DRAFT` pages in the menu, e.g. https://www.researchobject.org/ro-crate/1.2-DRAFT/terminology.html - however these still appear in [search](https://www.researchobject.org/ro-crate/search.html?q=terminology&lang=en):

This search is implemented by [theme.js](https://github.com/rundocs/jekyll-rtd-theme/blob/v2.0.10/assets/js/theme.js) in jekyll-rtd-theme that looks at <https://www.researchobject.org/ro-crate/pages.json> (yes, that is all the pages' HTML in one big JSON file!).

The https://github.com/rundocs/jekyll-rtd-theme we use do not have an exclude mechanism, but we added `exclude:true` to the markdown of [`1.2-DRAFT/README.md`](https://github.com/ResearchObject/ro-crate/blob/master/docs/1.2-DRAFT/README.md) (which skipped the whole 1.2-DRAFT/ tree in the menu) and override [site_pages.liquid](https://github.com/ResearchObject/ro-crate/blob/master/docs/_includes/reset/site_pages.liquid#L30) to skip those.

However each of the subpages still appear in that `pages.json`.

I am not sure if this issue might be solved just by adding `exclude:true` to each of the 1.2-DRAFT/* pages or if we would also need to modify `theme.js` in which case perhaps propose a pull request to jekyll-rtd-theme so it supports better excluded pages.

bedroesb:
I deployed the website in another theme giving us multiple advantages over the existing one:

- We can use the latest version of the theme
- Easy styling
- More modern look
- More flexibility in the side menu + links/ breadcrumbs always work
- you can easily exclude pages from the sidenav + ordering
- active community behind it
- automatic listing of subpages in 'parent' pages
- better search
- easy to exclude whole folders from sidenav and search in the config file (closes #123 )
- no advertisement of the theme

You can preview it [here](https://bedroesb.github.io/ro-crate/)

Reason I tried it because I felt really enclosed in the current theme. We have to do some hacky things in order to have simple things.

## The callouts look like this: 

Screenshot from 2021-03-09 22-03-13

## images:

Screenshot from 2021-03-09 22-04-19

## Code snippets:

Screenshot from 2021-03-09 22-04-54


stain:
It's quite nice! Fits better our colours as well. And I remember @ljgarcia talking about a theme like this as well. 

However I think lack of in-page header navigation is a big loss, we would have to restructure to lots of little pages to use this style, but that discourages browsing. Have they got an issue for TOC support?

In this one we don't need as much to give RO-Crate 1.1 section different navigation style, as it opens and expands automatically depending on the page.

stain (changes requested):
This is very nice, and good that it has proper support for the warnings, excludes, subpage-listing etc - less hacks needed! `nav_order` also seem to allow re-ordering without exposing the numbers which is good - at least for main page.

Stylistically this fits better with the colours from the logo etc, and seems more readable for longer pieces of text. For the "main" home page section also this will look less techy than the current theme, a bit more professional.

However I really miss the in-page header navigation which I felt was a big improvement by the current theme. I guess as a minimum we could add `{:toc}` or equivalent on top of each page? See https://bedroesb.github.io/ro-crate/1.0/ for one-toc-on-a-large-page.

Perhaps also explore how we can port over that functionality from the old theme or at least raise an issue upstream for this new theme?

https://bedroesb.github.io/ro-crate/1.2-DRAFT/ index is still hidden - with `nav_exclude` it should not need `_README` hack anymore.

bedroesb:
@stain good news, table of contents are possible: https://pmarsceill.github.io/just-the-docs/docs/ui-components/buttons/ 
we need to use the {:toc} snippet. But it would indeed be nice if it showed up as a sticky menu: a bit like this https://getbootstrap.com/docs/5.0/getting-started/introduction/ but on the right. Would fit the theme nicely imo.

Small disclaimer: I am now linking to a feature branch because the warnings are net yet in the main theme, but in a PR.

stain:
I guess we can use https://pmarsceill.github.io/just-the-docs/docs/navigation-structure/#in-page-navigation-with-table-of-contents approach for now.

Approach on that [bootstrap page](https://getbootstrap.com/docs/5.0/getting-started/introduction/) is confusing because it has two TOCs.. which I guess was easier to implement because `{:toc}` mechanism is done at markdown level while page-iteration is done at jekyll level.

bedroesb:
@stain in the different versions of the specifications, you sometimes have a README and an about, which one can become the index page for that specification? and do we both need them?
What was {% include references.liquid %} originally doing? Because that one is not working, but I also can't see the result of it in the current theme.

bedroesb:
@stain I don't know why the favicon does not work, because if you click in the HEAD of the page on the link towards it, he finds it ...

bedroesb:
besides that and the redirect it is ready I think :)

stain:
You rename `in-use/` to `using_rocrate` which I think makes less nice URIs - and I guess causes the conflict on `life_monitor.md`?

Which are the redirects needed, just `1.1/about` -> `1.1/`  and `implementations/` to `tools/` perhaps?

stain:
Agree  we don't need `README` as well as `about` as in this theme the index page will be clickable from navigation menu.

stain (approved):

bedroesb:
@stain I agree, the in-use is better. I don't know why I did that. About the redirects, could you have a look in making them work again?

bedroesb:
@stain I think we can merge :)

bedroesb:
wait I will quickly do the readme and about redirects

PR #140, opened 2021-03-15
Fixing title of index page

bedroesb:
Title of [index page ](https://www.researchobject.org/ro-crate/) is ` - Research Object Crate (RO-Crate)`

and adding back to top button

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -1,6 +1,5 @@
 remote_theme: pdmosses/just-the-docs@callouts
 logo: assets/img/ro-crate-wide.svg?20201023T1353
-#baseurl: https://www.researchobject.org/ro-crate/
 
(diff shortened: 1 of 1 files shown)


bedroesb:
@stain I don't know if this is worth the try?

stain:
Let's see what happens - just before the call!

bedroesb:
@stain damn it, now the title is 
```
<title>Research Object Crate (RO-Crate) | RO-Crate is a community effort to establish a lightweight approach to packaging research data with their metadata. It is based on schema.org annotations in JSON-LD, and aims to make best-practice in formal metadata description accessible and practical for use in a wider variety of situations, from an individual researcher working with a folder of data, to large data-intensive computational research environments.</title>`
``

PR #142, opened 2021-03-25
Second try-out for the index page title

bedroesb:
It could have been because title was in capital!


bedroesb:
@stain Since tools and use-cases are in a directory with their own index file, we can not refer to them as tools.md/.html but rather tools.

stain (approved):

PR #144, opened 2021-04-01
Fixing typos

marc-portier:
removing odd --> trailer that showed up in the pdf rendering
as well as double 'of' occurrence

### DIFF
--- a/docs/1.2-DRAFT/contextual-entities.md
+++ b/docs/1.2-DRAFT/contextual-entities.md
@@ -446,12 +446,10 @@ As the RO-Crate uses _flattened_ JSON-LD, `sdLicense` should be expressed direct
 
 **Tip**: If metadata is imported from a source licensed as [CC0 Public Domain Dedication][CC0], no `sdLicense` statement is required.
 
--->
-
 
 ## Extra metadata such as Exif
 
-Schema.org has a generic extension mechanism for encoding arbitrary properties and values which are not available as Schema.org properties. An example of of this is the Schema.org [recommended way (see example 2)][ImageObject] of including [Exif] technical image metadata.
+Schema.org has a generic extension mechanism for encoding arbitrary properties and values which are not available as Schema.org properties. An example of this is the Schema.org [recommended way (see example 2)][ImageObject] of including [Exif] technical image metadata.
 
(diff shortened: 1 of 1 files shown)


stain (approved):

stain (approved):

stain:
The `-->` is the end of a large `<!-- This got a bit to complicated` comment disabling the **License of contextual entity metadata** sub-section. Perhaps we should just remove that section we didn't re-add it in 1.1 either.

Reproduced here for longevity:


`<!-- TODO: This got a bit to complicated, commented out for 1.0 `

## License of contextual entity metadata

In some cases, the JSON-LD metadata for some entities have been imported under a different (possibly more restrictive) license than the license of the _RO-Crate Metadata File Descriptor_ overall. For this the property [sdLicense] ("structured data license") MAY be used on the affected data entities or contextual entities. In this case it is RECOMMENDED to use [sdPublisher] ("structured data publisher") for attribution of the imported metadata:

```json
{
  "@id": "./",
  "@type": "Dataset",
  "license": {
    "@id": "https://www.gnu.org/licenses/gpl-3.0"
  },
  "contentLocation": {
    "@id": "http://sws.geonames.org/8152662/"
  } 
},
{
  "@id": "http://sws.geonames.org/8152662/",
  "@type": "Place",
  "sdLicense": {
    "@id": "https://creativecommons.org/licenses/by/4.0/"
  },
  "sdPublisher": {
    "@id": "http://www.geonames.org"
  },
  "http://www.geonames.org/ontology#countryCode": "AU",
  "http://www.geonames.org/ontology#wikipediaArticle": {
    "@id": "https://en.wikipedia.org/wiki/Catalina_Park"
  },
  ...
}
```

In the above (abridged) example, there is no explicit license on the _RO-Crate Metadata File Description_, so the _Root Data Entity_ license [GPL 3.0] would apply to RO-Crate JSON-LD statements, except for the statements on the imported <http://sws.geonames.org/8152662/>,  which metadata is re-distributed under license <https://creativecommons.org/licenses/by/4.0/>.

In this example the CC-BY license requires retaining "a notice that refers to this Public License" and "identification of the creator(s) of the Licensed Material", here respected using `sdLicense` and `sdPublisher`.

As the RO-Crate uses _flattened_ JSON-LD, `sdLicense` should be expressed directly on each data/contextual entities where required.

**Tip**: If metadata is imported from a source licensed as [CC0 Public Domain Dedication][CC0], no `sdLicense` statement is required.

`-->`

Issue #153, opened 2021-05-26
Use Case: Indicate profile of RO-Crate


stain:
As a **specializer of RO-Crate**, I want to **indicate a particular RO-Crate profile** so that **consumers can make programmatic assumptions**.

For instance, an RO-Crate might follow the profile <https://about.workflowhub.eu/Workflow-RO-Crate/> but at the moment this can only be inferred by the right elements being present. Ideally one could also be *explicit* about which profile the crate intends to follow.

Given such a statement a programmatic consumer of RO-Crates can assume certain _duck typing_ as recommended by the profile - for instance that all `Person`s should have an `affiliation` to an `Organization` with a `url` property.

Formally such profiles might even include RDF shapes or similar to do such validation, but as a base case, as discussed in [@marc-portier's RO-Crate Profiles](https://docs.google.com/document/d/1-H6uhKQn-CFaB3WDwgqgQ6fJtQFNQzDA9VxEQq96kxc/edit#heading=h.2hfql8r6sor0) and [community call 2021-05-13](https://s.apache.org/ro-crate-minutes#heading=h.bsymzq245e1g) we want to just indicate the profile by URI.

marc-portier:
1/ 
Like the new idea where the formal-profile-description ends up being a crate itself - might be useful to hint towards an example  showing that off ? --> i.e. showing the proper encoding-format, hinting further towards a possible organization of additional stuff going into the profile?

2/
Reminding that we did also raise the topic of multiple json-ld contexts --> in issue #150 as a side effect of the profile discussion, I think it would be useful to have the resulting two additions to the spec refer to each other (when they land) ?

3/
Finally: some expressed use cases for having profiles in the first place (content structuring, validation, ...) have not been addressed yet - is the idea to:
a/ address those in separate issues, and have this issue strictly focused on the basics (purl, conforms-to, ...) 
b/ go as far as even introducing a 'ro-crate-profile-profile' leaving some template and additional semantics that structures, orders and covers all those (optional) elements
c/ not cover these aspects as part of the spec, but providing it as examples of profile-use?
d/ entirely leave it as an exercise to the reader, just provide a managed link-base of existing profiles?

(not trying to look too far into the future here: things always change at some point, no way we can predict any of that --  I'm more questioning what our current ambitions are / should be)

stain:
#154 was merged - see https://www.researchobject.org/ro-crate/1.2-DRAFT/profiles.html

A few outstanding issues:

[...]

PR #154, opened 2021-05-26
RO-Crate Profile section

stain:
See issue #153. 

This adds a section to explain:
* What is an RO-Crate profile? (borrowed from https://www.researchobject.org/ro-crate/profiles.html)
* How to assign an RO-Crate profile
* How to type a formal RO-Crate profile (e.g. SHACL RDF Shape)
* How to assign profiles to other entities

It does *not* (yet) detail the [suggestions on how to formally specify an RO-Crate profile](https://docs.google.com/document/d/1-H6uhKQn-CFaB3WDwgqgQ6fJtQFNQzDA9VxEQq96kxc/edit#heading=h.xhuk5kjevxx2) as a new or adapted expression language. 

This proposal have multiple sections we may not want to all include:

* [Declaring an RO-Crate profile](https://github.com/ResearchObject/ro-crate/blob/profiles/docs/1.2-DRAFT/profiles.md#declaring-an-ro-crate-profile) – expanding `conformsTo` on Metadata File descriptor
* [Which type of profile?](https://github.com/ResearchObject/ro-crate/blob/profiles/docs/1.2-DRAFT/profiles.md#which-type-of-profile) – indicating `encodingFormat` for formal profiles
* [Declaring a profile's profile](https://github.com/ResearchObject/ro-crate/blob/profiles/docs/1.2-DRAFT/profiles.md#declaring-a-profiles-profile) –  – being more specific than `encodingFormat` - specifically to indicate the use of SHACL inside RDF Turtle.
* [Declaring an entity's profile](https://github.com/ResearchObject/ro-crate/blob/profiles/docs/1.2-DRAFT/profiles.md#declaring-an-entitys-profile) – showing how profiles can be applied to other data entities like a Word document
* [Declaring an entity's metadata profile](https://github.com/ResearchObject/ro-crate/blob/profiles/docs/1.2-DRAFT/profiles.md#declaring-an-entitys-metadata-profile) – adding a new `sdConformsTo` to show that the JSON-LD entity conforms
  - Full example with a workflow shows distinction between `@type`, `sdConformsTo`, `encodingFormat`, `conformsTo` and `programmingLanguage`
* [Combining formal profiles using RO-Crate](https://github.com/ResearchObject/ro-crate/blob/profiles/docs/1.2-DRAFT/profiles.md#combining-formal-profiles-using-ro-crate) – suggesting an RO-Crate itself can be a profile to aggregate multiple schemas etc. (not expanded to example)

Now I added all these sections mainly for discussion – I am not convinced we want all of them! And perhaps some should move to [data entity section on encoding](https://www.researchobject.org/ro-crate/1.1/data-entities.html#adding-detailed-descriptions-of-encodings) or at least be linked to from there. XML is a tricky one because there are many declared subformats of XML but even more undeclared ones.

I added `sdConformsTo` in [ro-terms](https://github.com/ResearchObject/ro-terms/blob/master/vocabulary.csv) as I realised https://www.researchobject.org/ro-crate/1.1/workflows.html#complying-with-bioschemas-computational-workflow-profile had then wrong use of `conformsTo` (the Knime file content does not conform!) - this could however be simplified by just lifting those profiles up to also be `conformsTo` on the `ro-crate-metadata.json`

### DIFF
--- a/docs/1.2-DRAFT/context.jsonld
+++ b/docs/1.2-DRAFT/context.jsonld
@@ -2638,6 +2638,7 @@
           "retrievedOn": "http://purl.org/pav/retrievedOn",
           "retrievedBy": "http://purl.org/pav/retrievedBy",
           "conformsTo": "http://purl.org/dc/terms/conformsTo",
+          "sdConformsTo": "https://w3id.org/ro/terms#sdConformsTo",
           "@label": "http://www.w3.org/2000/01/rdf-schema#label",
           "pcdm": "http://pcdm.org/models#",
           "bibo": "http://purl.org/ontology/bibo/",
--- a/docs/1.2-DRAFT/metadata.md
+++ b/docs/1.2-DRAFT/metadata.md
@@ -117,6 +117,11 @@ These terms are being proposed by [Bioschemas profile ComputationalWorkflow 0.5-
 {: .note }
 > In this specification the proposed Bioschemas terms use the temporary <https://bioschemas.org/> namespace; future releases of RO-Crate may reflect mapping to the <http://schema.org/> namespace.
 
+Additional terms in RO-Crate's own [ro-terms vocabulary][ro-terms]:
+
+* `sdConformsTo` mapped to <https://w3id.org/ro/terms#sdConformsTo> (see [profiles](profiles.md))
+
+
 ## Summary of Coverage
 
 RO-Crate is simply a way to make metadata assertions about a set of files and folders that make up a _Dataset_. These assertions can be made at two levels:
--- a/docs/1.2-DRAFT/profiles.md
+++ b/docs/1.2-DRAFT/profiles.md
@@ -1,0 +1,107 @@
+---
+title: Profiles
+excerpt: |
+  Profiles of RO-Crates and their entities can be used to specialize and recommend
+  further which metadata types and properties are expected.
+nav_order: 9
+parent: RO-Crate 1.2-DRAFT 
+---
+    
+While RO-Crates can be considered general-purpose containers of arbitrary data and open-ended metadata, in practical use within a particular domain, application or framework, it will be beneficial to further constrain RO-Crate to a specific **profile**: a set of conventions, types and properties that one minimally can require and expect to be present in that subset of RO-Crates.
+
+Defining and conforming to such a profile enables reliable programmatic consumption of an RO-Crate’s content, as well as consistent creation, e.g. a form in a user interface form firmly suggest the required types and properties, and likewise a rendering of an RO-Crate can easier make rich UI components if it can reliably assume for instance that the [`Person`](contextual-entities.md#people) always has an `affiliation` to a [`Organization`](contextual-entities.md#organizations-as-values) which has a `url` - a restriction that may not be appropriate for all types of RO-Crates.
+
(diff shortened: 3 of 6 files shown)


stain:
See also schemaorg/schemaorg#2887 on adding `structuredData` (this metadata object)  and schemaorg/schemaorg#1516 on adding `conformsTo` to schema.org

stain:
Following [profile taskforce meeting 2021-06-18](https://docs.google.com/document/d/1Sw7WsMfcID1hxL1nKmXBAUUQlpgTl1ywrTNgLwX1nQs/edit#) it was decided to go for a simpler approach of:

1. RO-Crate profiles are indicated by expanding `conformsTo` on `ro-crate-metadata.json`
2. The profile indicated by `conformsTo` should be a `WebPage` that gives human readable information
3. The profile MAY optionally also resolve to another RO-Crate, a _Profile Crate_
4. The Profile Crate will have `hasPart` whatever schemas, documentation, software are relevant to this profile (but must still have a human-readable description)
5. The Profile Crate is resolved either by Content Negotiation or by adding `./ro-crate-metadata.json` to URL.

I have simplified the [draft](https://github.com/ResearchObject/ro-crate/blob/profiles/docs/1.2-DRAFT/profiles.md) to reflect the above.  I have also changed how profile languages are expressed: rather than a nested `conformsTo` (which may be confusing here), I used [encodingFormat](data-entities.md#adding-detailed-descriptions-of-encodings) URIs as we already suggest for other entities.

stain:
I also added a suggestion on how to indicate [extension vocabularies](https://github.com/ResearchObject/ro-crate/blob/profiles/docs/1.2-DRAFT/profiles.md#extension-vocabularies) and JSON-LD contexts.

Here I used <http://schema.org/DefinedTermSet> and <http://schema.org/DefinedTerm> which seems quite appropriate - perhaps we should also use those instead of `Thing` in <https://www.researchobject.org/ro-crate/1.1/appendix/jsonld.html#extending-ro-crate> ?

stain:
I tried realizing this as an example from the Workflow RO-Crate profile:

https://about.workflowhub.eu/Workflow-RO-Crate/ro-crate-preview.html
 
Notice how this was retrofitted to the old URL `https://about.workflowhub.eu/Workflow-RO-Crate/` without change – ideally this should of course also be versioned and so on.

Example shows most of the features, and adds a list of `licenses` that Workflow Hub expects.

stain:
Agreed (somewhat) in [RO-Cate community call 2021-07-08](https://s.apache.org/ro-crate-minutes#heading=h.tpmd0i1tmknk) to merge as-is for _1.2-DRAFT_, but also:

- [ ] Add diagram
- [ ] Explore further semantics, inheritance, templates and MUST/SHOULD/MAY requirements

marc-portier:
@stain - with some minor suggestions (scrible? expectations? format?) I could throw up a first draft for that diagram ...

stain:
Thanks, Marc! 

For the diagram, perhaps something like:

```
Example RO-Crate —conformsTo→ PID —redirect→ WFRun ProfileCrate —hasPart→ 
    - profile-for-describo.json —encodingFormat→ DescriboJSONProfile
    - ComputationalWorkflow
    - licenses/
```

but using some icons. Also not sure if the ProfileCrate itself should say it is a profile crate (`conformsTo: https://www.researchobject.org/ro-crate/1.2-DRAFT/profiles.html` or so?) or if that becomes a bit too much turtles all the way – at least explanation wise!

See https://about.workflowhub.eu/Workflow-RO-Crate/ro-crate-preview.html and https://about.workflowhub.eu/Workflow-RO-Crate/example/ as example.

PR #159, opened 2021-07-07
WIP: Front page in Bootstrap 5?

stain:
A work in progress of making a front page in Bootstrap 5.

https://raw.githack.com/ResearchObject/ro-crate/front/docs/front.html

Basically I've used https://getbootstrap.com/docs/5.0/examples/heroes/ but picked the wrong style for each picture.

Could someone help place this out and make it look almost pretty?


stain:
From [RO-Crate meeting 2021-07-08](https://s.apache.org/ro-crate-minutes#heading=h.v0kl3mtju7i6):

- [ ] @ptsefton to verify re-licensing of images to Apache License 2.0.
- [ ] @stain to fix font-linebreak problem in SVG.
- [ ] @marc-portier to have a go at rejigging the HTML/Bootstrap

Still undecided: Will this be the new index.html (splash screen-like with own style), the new introduction (embedded in same menu) or separate page linked to from a jumbatron-thing?

--> For now work with it as a separate so we are not constrained by the current theme.

marc-portier:
attempt at new styling to be found in 
https://github.com/ResearchObject/ro-crate/commit/c65b34d57dfa67103ef71c57f10d823423d51a65

can't say I know how this raw-githack trick works and if I should deliver a similar preview-support?


todo
- get feedback and not ignore it :)
- best review my introduced img/@alt messages too
- need more svg images to include --> happy to harvest those but could us a hint towards the source

marc-portier:
extra morning thought:

maybe smart to have each (or some) of these "headlines" link with some 'read more..." into a specific section of the spec or the samples?

wdyt?
- can we suggest smart entry points for each?
- should I work up some template style for those too?

stain:
Looks much better already!

Agree, there should be a button or something for each.. either explanation or example. Probably not straight deep into the spec from the front page.

Notice that the paragraphs in the second half to a degree repeats the first blocks – and I am not sure we should have a picture for each of them. Perhaps these "arguments" could be in two text columns - so they come side by side.

marc-portier:
/me likes the images though (both the style as well as having them)  
I was hoping @ptsefton has more (available or in the working) we could be adding?

also like the simple repetition in the current left-right-alternating style
that will break if we start doing side-by-side text at some point

so if we end up lacking images, I would rather propose to
* add also a text-centering template (for the no image case)
* and then regularly cycle through the 3 templates: left-right-center

stain:
Agree on the images and keeping a consistent style – we can find/make more pictures! Let's just check for open license on anything 'stock' – if it's CC-BY we can add a tiny footer on that very page with attributions.

The alternative of cycling through designs as you said could also work well.

docs/front.html
```diff
@@ -0,0 +1,382 @@
...
+        border-width: 1px 0;
+        box-shadow: inset 0 .5em 1.5em rgba(0, 0, 0, .1), inset 0 .125em .5em rgba(0, 0, 0, .15);
+      }
+    </style>
+
+  </head>
+  <body>
+    <main>
+      
+  <div class="jumbotron jumbotron-fluid m-2">
```
bedroesb, line 30: I thought the jumbotron class was dropped in bootstrap 5?

stain: Right, https://getbootstrap.com/docs/5.0/examples/jumbotron/ does this 'manually' now.

bedroesb:
@stain I can make a PR on this one integrating the jumbotrons in the current theme?

bedroesb:
Todo: adding more images

stain:
This looks very nice now (see below)! As @bedroesb points out we just need more images.

image

alaninmcr:
I disagree. I think it is far too long and should be broken up into smaller pages

bedroesb:
If they can be grouped we could put those groups in a carousel. Just an idea.

marc-portier:
> I disagree. I think it is far too long and should be broken up into smaller pages

Depends on how you look at it?

**visually long**

We shouldn't worry on this IMHO

This 'ĺong' screenshot is not how people will experience this...

I think that "breaking up / paging" is exactly what you browser-view-port will do for you with the help of a scroll-bar?
If you take that into account, the remaining question is: Do you rather click to the next page, or just scroll down? (I prefer the latter, and I would argue that modern responsive design / mobile device targeting is assuming that idiom all over) 

In a similar way the carousel decision can be narrowed down to: do you want the scroll movement to appear up-down or left-right?  (/me would just stick to scrolling down)


**long content**

But the length of a 5' intro or elevator pitch - should really keep it's promise of 5 minutes
We currently have ended on 15 topics to cover - leaving us 20 seconds per topic

So I really think we have to trim the longer texts even further - and indeed help out with more self-explaining images
Our first 5 topics set the good example and pace, the rest should follow suit?

Also: each of these topics deserves a link through to more information or direct pointer to the relevant section of the spec.  I think that is the other way to help trim the content here, as well as connect to "the other pages we already have"

stain:
Closing this as https://www.researchobject.org/ro-crate-2024/ is a different approach in different style.

PR #164, opened 2021-07-17
Correct namespace for rdfs:comment

stain:
Also clarify that we currently we mean rdfs:Class/rdf:Property and not http://schema.org/Class and http://schema.org/Property – but perhaps those would be better?


stain:
Would this typo be warranted also for a 1.1.2 update?

stain:
We should include the typo fix in 1.1.2 even if we go for a different way to do properties in 1.2 (e.g. `DefinedTerm`)

stain:
Uups, merged this to late, so didn't make for the PDF version in Zenodo.

PR #165, opened 2021-07-26
Integrating the new front page in current theme

bedroesb:
- [x] Moving current index page to about
- [x] changing order of RO-crae 1.1 index page
- [x] Making new index page 
- [x] Moving bootstrap classes to the custom class file

Screenshot from 2021-07-26 15-49-45


stain:
Thank you! @marc-portier agreed to have a look at changing this into the Jekyll template so we can put the text into YAML without explicit `<div class="bootstrap">` - but would need to customize/override the remote theme.

PR #171, opened 2021-09-17
Ro in use expanded

stuzart:
improved and extended the descriptions of RO-Crate in use, mentioned it is also being used for Data, and linked to from the front page

### DIFF
--- a/docs/in-use/index.md
+++ b/docs/in-use/index.md
@@ -23,13 +23,44 @@ redirect_from:
 
 # RO-Crate in use
 
-These applications use or expose RO-Crates:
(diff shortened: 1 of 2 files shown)


stain (approved):
This is brilliant, Stuart, thank you!

Perhaps @ptsefton would double-check the arkisto list of ocfl-tools etc? Did we miss something?

simleo:
Thanks Stuart. [I've updated](https://github.com/ResearchObject/ro-crate/pull/171/commits/956315a83f88f883fbb12a1cbc9408023ccde328) the LifeMonitor spelling to the single-word version we're using now.

stain:
Will merge for now - @ptsefton can edit afterwards.

PR #173, opened 2021-09-30
added details about the WfExS use case

stuzart:
the text provided by @lrodrin , thanks!


lrodrin (commented):
The WfExS description is OK.

Thanks, @stuzart!

PR #180, opened 2021-11-12
Wrong knime file in the Complete Workflow Example

rsirvent:
The hasPart points to a non-existing knime file in the RO-Crate.


stain (approved):
Thanks, `retropath.knime` seems to have snuck down from the example above.

stain (approved):

PR #188, opened 2022-01-25
Internal references for license and contextual entities

stain:
Make it clearer that other schema.org properties can be used on the RO-Crate Root data entity. 
Adds forward link to the sections on license and contextual entities.


stain:
I hope there's nothing controversial here, it also adds "Additional properties of _schema.org_ MAY be used". We need similar suggestions elsewhere.

Issue #183, opened 2021-12-06
Use Case: Allow RO-Crates to have root dataset IDs other than "./"


ptsefton:
As a **profile designer**, I want **to allow RO-Crates to have IDs that are not `./`** so that ** ro-crate-metadata can be served over an API**. 

"./" was originally a magic name until Stian came up with the metadata descriptor - the convention is no longer needed and is meanigless if the ro-crate metadata is not in a file on disk but is served via an API.

- See also  #178

simleo:
The spec already allows this (since version 1.1): the root data entity [SHOULD have an `@id` of `./`](https://www.researchobject.org/ro-crate/1.1/root-data-entity.html#ro-crate-metadata-file-descriptor). This is [not handled by ro-crate-py yet](https://github.com/ResearchObject/ro-crate-py/issues/34) though. Perhaps we want to get rid of this recommendation and write something like "legacy specs (until version 1.0) required the root data entity to have an `@id` of `./`; this is still recommended for RO-Crates stored as file system directories" instead.

stain:
@simleo however  [properties of the root data entity](https://www.researchobject.org/ro-crate/1.1/root-data-entity.html#direct-properties-of-the-root-data-entity) includes:

>@id: MUST end with `/` and SHOULD be the string `./`

I think we can soften that `MUST` to another `SHOULD` - athough then the _Root Data Entity_ is more of risk of not having anything to do with the [_root directory_](https://www.researchobject.org/ro-crate/1.1/structure.html)  which could get confusing. 

So do we need to clarify the terminology then? This entity then becomes the "RO-Crate" rather than of the RO-Crate Root (which is effectively how it is used). Note that the current _root_ terminology helped to avoid "The RO-Crate is made of the RO-Crate" loops

simleo:
Discussed this morning at the meeting. One of the main issues raised was that if the root data entity id is an arbitrary URI, then it's not clear what to do with relative paths of data entities. @dgarijo came up with an idea that many of us found great, myself included: state that relative paths are unsupported in data entities (behavior of implementation is undefined) if the root data entity id is not `./`. In other words, the root data entity id may be an arbitrary URI, but then all data entities must have an absolute URI as their id.

dgarijo:
@simleo thanks for the credit, but I think it was @stain and @ptsefton who proposed it. I just rephrased emphasizing the 2 use cases focus on  (i.e., leave out of scope relative paths if the root data entity id is not `./`)

[...]

PR #189, opened 2022-01-27
Add concept Detached RO-Crate

stain:
.. to support #183 my logical conclusion is that we need the concept of a _Detached RO-Crate_.

Suggest definition (from this pull request's [structure.md](https://github.com/ResearchObject/ro-crate/blob/issue-183-nonslash-root/docs/1.2-DRAFT/structure.md#types-of-ro-crate))

> There are two classes of RO-Crate detailed below:

**Regular RO-Crate**
: A crate that has a well-defined _RO-Crate Root directory_ and can carry an **explicit payload** of local _data entities_ as regular files (combined with _Web-based Data Entities_ where needed). This type of RO-Crate can be suitable for long-term preservation, transfer and publishing, as the _RO-Crate Metadata File_ is stored alongside the crate's payload. 

**Detached RO-Crate**
: A crate without a defined payload directory. In this kind of crate, all data references are absolute. This approach may be suitable for use with dynamic web service APIs and repositories that can't preserve file paths. As the data of these crates can only be _Web-based Data Entities_, the **payload is implicit** and must be preserved/transferred/archived independent of the _RO-Crate Metadata File_. 

See further [definition of detached RO-Crate](https://github.com/ResearchObject/ro-crate/blob/issue-183-nonslash-root/docs/1.2-DRAFT/structure.md#detached-ro-crate)

I think this is necessary because of #183 allowing `@id` to be any ID, as here proposed in new sub section [Root Data Entity identifier](https://github.com/ResearchObject/ro-crate/blob/issue-183-nonslash-root/docs/1.2-DRAFT/root-data-entity.md#root-data-entity-identifier) - then

> If the `@id` of the Root Data Entity is an absolute URI, the Crate SHOULD NOT contain data entities using relative URI references, but MAY contain Web-based Data Entities using absolute URIs.

And from that my logical conclusion is that the whole concept of "RO-Crate Root" and any relative URIs becomes ambigious and difficult if we no longer have `"@id: ./"` of the Root Dataset and the URI that serves `ro-crate-metadata.json` no longer is grounded in something similar to a folder.

I would hope for some discussion on this in the RO-Crate meeting today 2022-01-27.


stain:
From [RO-Crate meeting 2022-01-27](https://s.apache.org/ro-crate-minutes#heading=h.yo6j9hcp2hy6):

- [x] General agreement to keep the suggested split of all-relative or all-absolute. 
- [ ] Change terminology, “Packaged RO-Crate”? “Transferred RO-Crate”?  
  → Peter to review this PR, then revise intro text accordingly. 
-  [ ] How to write this back out?  
  → Add to JSON-LD appendix on how to convert between absolute and relative using `@base` tricks.

ptsefton:
I suggest we use the term "Attached RO-Crate". I'm not sure how to edit the PR...

stain:
I started drafting a section [Converting from attached to detached](https://github.com/ResearchObject/ro-crate/blob/issue-183-nonslash-root/docs/1.2-DRAFT/appendix/relative-uris.md#converting-from-attached-to-detached-ro-crate)

just wanted to check if we are OK with what comes out of the JSON-LD flattening:

```json
{
  "@context": [
    {"@base": "arcp://uuid,d6be5c9b-132a-4a93-9837-3e02e06c08e6/"},
    "https://w3id.org/ro/crate/1.1/context"
  ],
  "@graph": [
    {
      "@id": "https://about.workflowhub.eu/Workflow-RO-Crate/1.0/ro-crate-metadata.json",
      "@type": "CreativeWork",
      "conformsTo": {"@id": "https://w3id.org/ro/crate/1.1"},
      "about": {"@id": "https://about.workflowhub.eu/Workflow-RO-Crate/1.0/"},
      "creator": {"@id": "https://orcid.org/0000-0001-9842-9718"}
    },
    {
      "@id": "https://about.workflowhub.eu/Workflow-RO-Crate/1.0/",
      "@type": "Dataset",
      "hasPart": [
        { "@id": "https://about.workflowhub.eu/Workflow-RO-Crate/1.0/index.html"},
        { "@id": "https://about.workflowhub.eu/Workflow-RO-Crate/1.0/example/"},
      ],
      "name": "Workflow RO-Crate profile"
    },
  {
      "@id": "https://about.workflowhub.eu/Workflow-RO-Crate/1.0/ro-crate-metadata.json#include-ComputationalWorkflow",
      "@type": "Recommendation",
      "category": "MUST",
      "name": "Include Main Workflow",
      "itemReviewed": {
        "@id": "https://bioschemas.org/ComputationalWorkflow"
      }
    }
```

1. `ro-crate-metadata.json` now have absolute `@id` (breaks a SHOULD?)
2. The root dataset matches the public Web presence (but then why did we need Detached?)
3. `#local` identifiers become grounded in the original RO-Crate
4. The identifier of the corresponding attached RO-Crate is preserved in `@id`

For anything more "proper" I think you would need manual processing, e.g. manual deposit and rewrite of each data entity file, manual UUID for each contextual entity.

simleo:
I think we should recommend removing the `{"@base": "arcp://uuid,.../"}` from the converted output. It can be confusing (I was wondering what its purpose was until I read the section about converting to detached).

ptsefton:
Are the uuids intended to be unique? Cos people will copy and paste, or hardcode them into their crate.

Regarding attached crates can we do the deal with the relativity of paths using base: "./" or similar (or is that not allowed?) I know I have base: null in crates to stop JSON-LD libraries from messing with my paths - would have to refresh my memory

jmfernandez:
If you leave the arcp based UUID, you should add a small recipe about using python's arcp or how to generate in a couple of programming languages those UUIDs in the namespace of URLs.

```python
import uuid
the_url = 'https://example.org'
the_uuid = uuid.uuid5(uuid.NAMESPACE_URL, the_url)
# the_uuid.hex has the UUID string representation
```

```python
import arcp
the_arcp = arcp.arcp_location("http://example.com/data.zip", "/file.txt")
# the_arcp has the ARCP string representation
```

```python
import uuid
the_random_uuid = uuid.uuid4()
# the_random_uuid.hex has the UUID string representation
```

```python
import arcp
the_random_arcp = arcp.arcp_random()
# the_random_arcp has the ARCP string representation
```

ptsefton:
On reflection I don't think we need this distinction. I think we should look at providing clear info about how to use relative and absolute paths for various resources. 

Based on experience where we have implemented an API that uses the API URL as the @id but it is then not clear how to reconstitute a crate, I think that approach was a mistake. It might be better to go back to an approach where @ids are 

1.  Relative URIs To describe how resources would be or  laid out on disk as a set of relative paths with ./ for the root
2. Absolute URIs For URL addressable resources

For packaged crates-on-disk use @base: null with relative paths for data entities

For crates over an API use the [dcat:downloadURL](https://w3c.github.io/dxwg/dcat/#Property:distribution_download_url)  property on DataEntities for the place where you can get a file and as per (1) above make its @id the filename it should have relative to the root. and Identifier for IDs like DOIs.

ptsefton:
Further to my last comment @stain & @simleo. I think I have found a neat solution to the problem we were having with letting "@id" in for a File be a URI - how would you save it to disk and re-construct the relative path structure of a package?

Solution:  In RO-Crate Metadata Documents served from a service leave the @ids as relative paths but use DCAT accessUrl (to point to RO-Crate Metadata served over an API) and downloadURL for the actual datastream. We can then recommend that a process for reconstituting an RO-Crate by using the @id to create directories and write file contents.


I have [written this up](https://docs.google.com/document/d/1_UrbtzjlXOMXOCoV1bNTzEjOObn0cXgiBpbQ1gRBepI/edit#heading=h.tl58db6ml499) in the work I was doing on a new intro - this detail probably does not all belong in the intro though.

Here's a copy and paste from that Google doc.

>  an RO-Crate Metadata Document is served from a service use the following DCAT properties:
> 
> dcat:accessURL – RO-Crate Metadata Documents
> 
> dcat:downloadURL - Direct downloads of bitstreams (files)
> 
> Client software to construct RO-Crates SHOULD:
> Save the RO-Crate Metadata file into an empty create download directory 
> For Dataset that has a relative URI, make a subdirectory  with the same path as the Dataset id - eg /data/pictures
> For each File Entitiy fetch the datastream using downloadURI and write it to a relative path that corresponds to its “@id” (creating directories as needed, even if they are not described in the RO-Crate).
> 
> In the case where the RO-Crate metadata Document is being served from a service,
> { "@context": "https://w3id.org/ro/crate/1.1/context", 
>   "@graph": [
> 
>  {
>     "@type": "CreativeWork",
>     "@id": "ro-crate-metadata.json",
>     "conformsTo": {"@id": "https://w3id.org/ro/crate/2.0"},
>     "about": {"@id": "./"}
>  },  
>  {
>     "@id": "./",
>     "accessUrl": "https://example.com/ro-crate/api/crate/000001",
>     "@type": "Dataset",
>     "datePublished": "2022-02-01",
>     "name": "Example dataset for RO-Crate specification",
>     "description": "If this were real data it would contain a minute-by-minute rainfall readings for my weather gauge",             "license": "CC BY-NC-SA 3.0"
>     "hasPart": {"@id": "data.csv"}
>  }
> {
>     "@id": "data.csv",
>     "downloadURL": https://example.com/ro-crate/api/crate/000001?file=data.csv,
>     "@type": "File",
>     "[encodingFormat](https://schema.org/encodingFormat)": "text/csv",
>     "name": "Rainfall data for Katoomba, NSW Australia, 2022-02-01",
>     "license": "CC BY-NC-SA 3.0 AU"
> 
>  }
> 
> ]
> }

jmfernandez:
Looks nice, @ptsefton @stain @simleo !! I have several questions, some of them offtopic.

* If the resource has a public id representation as a short URI using an scheme registered either at identifiers.org or at n2t.net in CURIE format, where should we put those public identifiers, as we are keeping `@id` to represent the internal, relative placement of the resource?
* If the resource has a download URL which is not http based (for instance, an FTP or an Amazon S3 link), is still `downloadURL` predicate the recommended one to provide it?
* Even when it is not possible to provide a `downloadURL`, is there some proper way to declare that the resource is under controlled access? For instance, data from EGA (European Genome Phenome Archive) or dbGaP (NCBI's database of Genomes and Phenomes)

simleo:
> I think I have found a neat solution to the problem we were having with letting "@id" in for a File be a URI - how would you save it to disk and re-construct the relative path structure of a package?

The current spec already allows `File` `@id`s to be URIs. In ro-crate-py, this is handled via a `fetch_remote` keyword argument that allows the library user to decide what happens when the crate is written out to disk (the user might not want to download the file -- or be able to do so -- for various reasons):

```python
url = "http://example.com/foo.txt"
# Download file; it will be placed under <CRATE_DIR>/examples when the crate is written out
crate.add_file(url, "examples/foo.txt", fetch_remote=True)
# Don't download file; its @id will still be a URI in the output crate
crate.add_file(url, fetch_remote=False)
```

In the latter case, a `"url": "http://example.com/foo.txt"` is automatically added to the entity; however, we're currently not doing that in the former case, and I now realize that we should. But maybe we should use "downloadUrl" rather than "url".

@jmfernandez I don't think there's any requirement for `URL` schemes to be `http[s]` in Schema.org.

docs/1.2-DRAFT/root-data-entity.md
```diff
@@ -44,7 +44,7 @@ The _RO-Crate JSON-LD_ MUST contain a self-describing
 **RO-Crate Metadata File Descriptor** with
 the `@id` value `ro-crate-metadata.json` (or `ro-crate-metadata.jsonld` in legacy
 crates) and `@type` [CreativeWork]. This descriptor MUST have an [about]
```
simleo, line 46: We need to clarify that the descriptor's id can also be an absolute URI. For instance, by add a sentence here like:

```
In a [detached RO-Crate](structure.md#detached-ro-crate), the descriptor's `@id` can be an absolute URI; in this case, its last path segment MUST be `ro-crate-metadata.json` (or `ro-crate-metadata.jsonld` in legacy crates)
```

simleo: An alternative is rephrasing the previous sentence as:

```
The _RO-Crate JSON-LD_ MUST contain a self-describing **RO-Crate Metadata File Descriptor**
whose `@id` MUST have `ro-crate-metadata.json` (or `ro-crate-metadata.jsonld` in legacy crates)
as its last path segment.
```

stain:
> UPDATE: https://schema.org/downloadUrl is only used in SoftwareApplication, so I guess we should use `url`

@simleo `url` is how we've said in https://www.researchobject.org/ro-crate/1.1/data-entities.html#embedded-data-entities-that-are-also-on-the-web - but how about https://schema.org/contentUrl which is defined for `MediaObject` aka our `File`, which would be the schema-org way of doing [dcat:downloadURL](https://w3c.github.io/dxwg/dcat/#Property:distribution_download_url) - rather than `url` which is just "URL of the item" (and therefore weird to understand when it's different from `@id`)

stain:
[Call 2023-03-23](https://docs.google.com/document/d/1vg805CJ4QXY6CCBNR-8ETgkbOCAMqSKJPCKOTNxrIbk/edit#heading=h.5rlou4obpiov) agreed to merge all outstanding PRs.

There's outstanding how to do re-construct the relative path -- @simleo may have also thoughts on this now from Workflow Run profile perspective which also needed to this.

PR #198, opened 2022-04-06
Update docs on finding the root dataset

simleo:
Updates the root data entity section as follows:

* Adds a paragraph to the metadata file descriptor spec to allow absolute URIs (e.g., `https://example.org/crate/ro-crate-metadata.json`) when the root data entity is also an absolute URI.
* Changes the recommended algorithm for finding the root data entity to leverage the constraint that the metadata MUST contain an `ro-crate-metadata.json` (or absolute URI whose last path segment is `ro-crate-metadata.json`) entry. This makes the algorithm simpler and more efficient in the "standard" case of a single entry with an `@id` of `ro-crate-metadata.json` (see https://github.com/ResearchObject/ro-crate-py/pull/119).

Note that, in order to make this PR independent from #189, I've avoided using the term "detached crate". If we merge that PR first, however, we should change this one to explicitly mention detached crates.

### DIFF
--- a/docs/1.2-DRAFT/appendix/relative-uris.md
+++ b/docs/1.2-DRAFT/appendix/relative-uris.md
@@ -308,30 +308,13 @@ When parsing _RO-Crate JSON-LD_ as RDF, where the RDF framework performs resolut
 The algoritm proposed in section [Root Data Entity](../root-data-entity.md#finding-the-root-data-entity) allows finding the RDF resource describing `ro-crate-metadata.json`, independent of its parsed base URI. We can adopt this for RDF triples, thus finding crates conforming to this specification can be queried with [SPARQL]:
 
 ```sparql
-PREFIX dcterms: <http://purl.org/dc/terms/>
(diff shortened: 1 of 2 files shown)


simleo:
Re: the nested crate example, it would be nice to recommend adding a `conformsTo` to the relevant entry in order to indicate that it's an RO-Crate:

```json
    {
      "@id": "http://example.org/crate/nested/",
      "@type": "Dataset",
      "conformsTo": {"@id": "https://w3id.org/ro/crate/1.2-DRAFT"}
    }
```

I haven't included that in this PR since in the current spec the `conformsTo` applies to the metadata file descriptor. However, we're planning to move it to the crate, so this is something to postpone to after we've done that.

stain:
We agreed that implementation can try to be lax and rescue user -- but spec should be more strict. Don't suggest the heuristics. Just note to not expose internals of nested crates as it could confuse root dataset algorithm.

ptsefton:
@stian where is this up to?

stain:

Issue #124, opened 2021-01-11
Use Case: Indicate main data entity of RO-Crate


stain:
As a **framework developer**, I want to **generate an RO-Crate** to wrap **one main data entity** with its associated resources, so that I can programmatically access the main entity and navigate its metadata.

For example, the [Workflow RO-Crate profile](https://about.workflowhub.eu/Workflow-RO-Crate) needs to indicate the main `ComputationalWorkflow` data entity, as there could be more than one in such specialized RO-Crates.

We can think of this as the "entry point" - if you don't know which data entity to look at first, start here.

We have so far used http://schema.org/mainEntity for this:

```json
    {
      "@id": "./",
      "@type": "Dataset",
      "name": "Example Workflow",
      "description": "An example workflow RO Crate",
      "license": "Apache-2.0",
      "mainEntity": {
        "@id": "example_workflow.cwl"
      },
      "hasPart": [
        {
          "@id": "example_workflow.cwl"
        },
        {
          "@id": "diagram.svg"
        },
        {
          "@id": "README.md"
        }
      ]
    }
```

In this example there are three data entities, but only `example_workflow.cwl` has been promoted to the `mainEntity`.

`mainEntity` is a specialization of https://schema.org/about  in that the expectation is that there is only one. Perhaps in our recommendation we also want to RECOMMEND that the main entity is a data entity also listed under `hasPart`, while `about` can have (multiple) contextual entities like a `Person` or `Place`.

See also #122

stain:
Can be contextual or data entity. If it's a contextual entity should it also be listed under `mentions`? #122 

Design choice of when are contextual entities listed under:
- `mentions`
- `about`
- `mainEntity`
- Reached from properties of other listed entities

In Workflow Run Crate the `mainEntity` stays as the workflow definition (as it inherits Workflow Crate)  -- but semantically it is really the `CreateAction` (workflow run) that is the main entity there.

Issue #122, opened 2020-12-09
Use Case: RO-Crate as collection of contextual items


stain:
As a **developer that's extending RO-Crate**, I want **contextual items to be gathered** so that **the Dataset is non-empty even if there are no data entities**.

Here's the real [use case](https://github.com/crs4/life_monitor/wiki/Workflow-Testing-RO-Crate) (see also https://github.com/ResearchObject/ro-crate-py/pull/38) by @simleo:

```json
{
    "@context": [
        "https://w3id.org/ro/crate/1.1/context",
        {
            "TestSuite": "https://w3id.org/ro/terms/test#TestSuite",
            "TestInstance": "https://w3id.org/ro/terms/test#TestInstance",
            "TestService": "https://w3id.org/ro/terms/test#TestService",
            "TestDefinition": "https://w3id.org/ro/terms/test#TestDefinition",
            "instance": "https://w3id.org/ro/terms/test#instance",
            "runsOn": "https://w3id.org/ro/terms/test#runsOn",
            "resource": "https://w3id.org/ro/terms/test#resource",
            "definition": "https://w3id.org/ro/terms/test#definition"
        }
    ],
    "@graph": [
        {
            "@id": "ro-crate-metadata.json",
            "@type": "CreativeWork",
            "about": {
                "@id": "./"
            },
            "conformsTo": {
                "@id": "https://w3id.org/ro/crate/1.1"
            }
        },
        {
            "@id": "./",
            "@type": "Dataset",
            "name": "sort-and-change-case",
            "description": "sort lines and change text to upper case",
            "license": "Apache-2.0",
            "mainEntity": {
                "@id": "sort-and-change-case.ga"
            },
            "hasPart": [
                {
                    "@id": "sort-and-change-case.ga"
                },
                {
                    "@id": "test/"
                },
                {
                    "@id": "LICENSE"
                },
                {
                    "@id": "README.md"
                },
                {
                    "@id": "test/test1/sort-and-change-case-test.yml"
                }
            ]
        },
        {
            "@id": "sort-and-change-case.ga",
            "@type": [
                "File",
                "SoftwareSourceCode",
                "ComputationalWorkflow"
            ],
            "programmingLanguage": {
                "@id": "#galaxy"
            },
            "name": "sort-and-change-case"
        },
        {
            "@id": "test/",
            "@type": "Dataset",
            "hasPart": [
                {
                    "@id": "#test1"
                }
            ]
        },
        {
[...]

PR #211, opened 2022-09-29
Attempting to address #122 and #124 - mainEntity and mentions

ptsefton:
I have drafted a new section which attempts to codify the discussions around mentions and mainEntity (I left out `about` as it is too hard to differentiate from `mentions` - this is complicated enough already)

I think this needs to be a short new section because while it belongs int he discussion of metadata on a root dataset it also needs to come after the sections on Data and Contextual entities.


stain:
[ ] Render to check new section is OK
[ ] Link to new section from relevant entity pages

stain:

PR #219, opened 2022-11-10
Update RO-Crate introduction with running example

stain:
This fixes #215 by integrating @ptsefton's new intro text.

Work in progress as figures are missing, and Markdown syntax needs a few more tweaks and links.

PR #220, opened 2022-11-10
Peter's draft on publishing RO-Crates

stain:
Work in progress. Several things to complete and consider here, so separating from #219.


stain:
Mentioned in meeting 2024-04-23, we don't think this is so relevant now, see #296 and

Issue #182, opened 2021-12-06
Use Case: Allow extra types on root dataset


ptsefton:
As a **profile designer** I want to have **additional types on the root dataset** to simplify writing profiles - the current use case is wanting to be able to double up with "RepositoryCollection" (pcdmCollection) or "RepositoryObject" (pcdm:Object) to distinguish between collections and "actual data".

ptsefton:
We're doing this now - does anyone object to it being made legit in the spec?

stain:
I think it should be allowed. This is also useful for profiles #153 if we want to type as [`prof:Profile`](https://www.w3.org/TR/dx-prof/) which seems a useful type for profiles together with profile nesting `prof:isProfileOf `

PR #223, opened 2022-11-23
Changed to text to allow multiple types on root - as per #182 which w…

ptsefton:
as agreed but not actioned


stain:
Might be confusing to introduce `creator` here as we've tried to be consistent with using `author`. If we recommend both then we should distinguish them, perhaps like in PAV [createdBy](http://pav-ontology.github.io/pav/#http://purl.org/pav/createdBy) vs [authoredBy](http://pav-ontology.github.io/pav/#http://purl.org/pav/authoredBy) as explained [in PAV paper](https://jbiomedsem.biomedcentral.com/articles/10.1186/2041-1480-4-37#Sec7)

docs/1.2-DRAFT/introduction.md
```diff
@@ -146,7 +146,7 @@ In RO-Crate Metadata Files, entities are cross-referenced using `@id` reference
 }
 ```
 
-The root is always typed `Dataset` and have several metadata properties that describe the RO-Crate as a whole, as a collection of resources. The section on [root data entity](root-data-entity.md) explores further the required and recommended properties of the root `./`.
+The root is always has `Dataset` as one of its values, though it may have more than one type. It has several metadata properties that describe the RO-Crate as a whole, as a collection of resources. The section on [root data entity](root-data-entity.md) explores further the required and recommended properties of the root `./`.
```
simleo, line 149: "is always has" :arrow_right:  "always has". Also, `Dataset` is one of its types, not values. I propose changing the sentence to:

"The root always has the `Dataset` type, though it may have more than one"

stain:
Merging as [call 2023-03-23](https://docs.google.com/document/d/1vg805CJ4QXY6CCBNR-8ETgkbOCAMqSKJPCKOTNxrIbk/edit#heading=h.5rlou4obpiov) agreed to merge all outstanding PR (that are not WIP) so we can do an overall review and push out 1.2.

PR #229, opened 2023-01-16
conformsTo profile applies to root dataset, not metadata file

### LINKED ISSUE #126: Use Case: As a funder I want to be able to see from the RO-Crate website that it is (a) being used (b) growing (c) has substance
I have started a document for us to discus the much-needed revamp of the website.

https://docs.google.com/document/d/1sMIoExD0Z_BvwjD0jwD4jrvZDz143p4MqG0C34EL_9o/edit?usp=sharing

@CaroleGoble Point taken that we need to improve our marketing would appreciate you putting in the goals you have for the site. I have started putting in some content about existing case studies - more needed from the work we've been hearing about in the RO-Crate calls.

stain:
To indicate RO-Crate following a particular [profile](https://www.researchobject.org/ro-crate/1.2-DRAFT/profiles), then move so "conformsTo" applied to "./" as suggested in https://github.com/ResearchObject/ro-crate/issues/153#issuecomment-1240446098 (and agreed in calls [2022-09-08](https://docs.google.com/document/d/1qBCqy7GXTPyydjDIC5h0VGS_4TDSzTnlIRg6dhjyAJ4/edit#heading=h.3a974lmk0gw0) and [2022-01-27](https://docs.google.com/document/d/1qBCqy7GXTPyydjDIC5h0VGS_4TDSzTnlIRg6dhjyAJ4/edit#heading=h.yo3tse270vuq)) rather than on `ro-crate-metadata.json` entity.

This means `conformsTo` on the `ro-crate-metadata.json` is only for RO-Crate spec itself (considering it from a JSON/JSON-LD perspective) while profiles about the content/shapes of a class of crates are appliec using `conformsTo` on the root dataset (`./`)

This also makes it easier to link to other RO-Crates as a `Dataset` and saying they follow a particular profile -- however you would also need to add a `subjectOf` to link to their `ro-crate-metadata.json` in order to indicate it is an RO-Crate - so still not ideal. 

(also there is still no way to indicate RO-Crate-ness on a nested `Dataset` which has a `distribution` to a Zip/tar `DataDownload` according to https://www.researchobject.org/ro-crate/1.1/data-entities.html#directories-on-the-web-dataset-distributions.)

See also ResearchObject/workflow-run-crate#40 which applies this on the Workflow Run Crate profiles.

This fixes #126

### DIFF
--- a/docs/1.2-DRAFT/profiles.md
+++ b/docs/1.2-DRAFT/profiles.md
@@ -106,7 +106,7 @@ A Profile Crate MUST declare a human-readable _profile description_, which is [a
     "@type": "File",
     "name": "Workflow RO-Crate profile description",
     "encodingFormat": "text/html",
-    "about": "./",
+    "about": "./"
 }
 ```
 
@@ -120,7 +120,7 @@ The _profile description_ MAY be equivalent to the
     "@type": "CreativeWork",
     "name": "RO-Crate preview of the Process Run Crate",
     "encodingFormat": "text/html",
-    "about": "./",
+    "about": "./"
 }
 ```
 
--- a/docs/1.2-DRAFT/profiles.md
+++ b/docs/1.2-DRAFT/profiles.md
@@ -61,9 +61,7 @@ RO-Crates conforming to (or intending to conform to) such a profile declare this
 {
     "@id": "./",
     "@type": "Dataset",
-    "conformsTo":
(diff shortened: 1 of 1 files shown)


simleo (approved):
I've changed the profile crate retrieval section to use Workflow RO-Crate for all examples. This way, since Workflow RO-Crate has a profile crate, readers can actually try following the instructions and check that they get JSON-LD when requested.

PR #230, opened 2023-01-16
Clarify properties for data entities, avoid WebSite type

### LINKED ISSUE #99: encodingFormat use CWL example, mismatch with Workflow section
**Describe the bug**

In https://www.researchobject.org/ro-crate/1.1-DRAFT/data-entities.html?highlight=commonwl#adding-detailed-descriptions-of-encodings example we see an example of `encodingFormat` being a URI rather than media type: 

```json
 {
    "@id": "1st-tool.cwl",
    "@type": "File",
    "contentSize": "120",
    "description": "An example Common Workflow Language File",
    "encodingFormat": ["text/plain", {"@id": "https://www.commonwl.org/v1.0/Workflow.html"}]
  },
    {
    "@id": "https://www.commonwl.org/v1.0/Workflow.html",
    "name": "Common Workflow Language (CWL) Workflow Description, v1.0.2",
    "@type": "Website"
  }
```

First of all (@mr-c might correct me) I think the better URI here would be CWL's permalink <https://w3id.org/cwl/v1.0/> as shown on <https://www.commonwl.org/v1.0/Workflow.html>

Secondly this use of `encodingFormat` here seems to be at odds with `programmingLanguage` in [Workflows section](https://www.researchobject.org/ro-crate/1.1-DRAFT/workflows.html#describing-scripts-and-workflows) - it is unclear which one of these should be used - or both?

**Suggested fix**

Easiest fix is to avoid mentioning workflows at the `encodingFormat` example - perhaps use a geopositional format or something instead.

Longer fix would be to reconcile the two, or at least mention `encodingFormat` under the Workflow section. Most programming languages don't have a registered [IANA media type](https://www.iana.org/assignments/media-types/) but would usually be a variant of `text/plain` (CWL is encoding format `text/x-yaml` - even YAML does not have a media type yet).

So perhaps <https://yaml.org/spec/1.1/> could be a better example, but generic YAML rather than CWL-that-happens-to-be-YAML.

stain:
Also make it clear that other `CreativeWork` properties can be used. The use of [WebSite](http://schema.org/WebSite) for file formats is replaced with [WebPage](http://schema.org/WebPage) as there is typically not a whole website just for one file format (and in particular the https://www.nationalarchives.gov.uk/PRONOM/fmt/19 file formats are part of the `nationalarchives.gov.uk` WebSite).

This also fixes #99 by replacing the CWL example with a Gromacs trajectory which has a special page.

### DIFF
--- a/docs/1.2-DRAFT/data-entities.md
+++ b/docs/1.2-DRAFT/data-entities.md
@@ -114,7 +114,7 @@ An example _RO-Crate JSON-LD_ for the above would be as follows:
 The above example provides a media type for the file `cp7glop.ai` - which is
 useful as it may not be apparent that the file is readable as a PDF file from the
 extension alone. To add more detail, encodings SHOULD be linked using a [PRONOM]
-identifier to a _Contextual Entity_ of `@type` [WebSite].
+identifier to a _Contextual Entity_ of `@type` [WebPage].
 
 ``` json
   {
@@ -128,7 +128,7 @@ identifier to a _Contextual Entity_ of `@type` [WebSite].
   {
     "@id": "https://www.nationalarchives.gov.uk/PRONOM/fmt/19",
     "name": "Acrobat PDF 1.5 - Portable Document Format",
-    "@type": "WebSite"
+    "@type": "WebPage"
   }
 
 ```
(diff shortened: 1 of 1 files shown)


stain:
Rendered at https://stain.github.io/ro-crate/1.2-DRAFT/data-entities

PR #232, opened 2023-01-17
Ad-hoc terms use DefinedTerm/Class/Property

stain:
http://schema.org/DefinedTerm is better than http://schema.org/Thing for pre-existing URIs that are semantic artefacts or concepts. Somewhat equivalent to [skos:Concept](https://www.w3.org/TR/skos-primer/).  This will then be consistent with https://www.researchobject.org/ro-crate/1.2-DRAFT/profiles that also mentions http://schema.org/DefinedTermSet.

For simplification with the profiles, `rdf:Property` / `rdfs:Class` types now optional, so we can stay in schema.org land for lightweight properties using their equivalent http://schema.org/Property and http://schema.org/Class. The RDFS variants are needed to do `rdfs:subPropertyOf` or `rdfs:subClassOf` but not I think needed for just adding a few terms to JSON-LD.

The part about `@context` and ad-hoc terms will still need to be updated to show that a RO-Crate Profile Crate could be used directly as a `@context` to use its terms there.

Also fixed some bugs in the expanded RDFS code from #208 (as `@id` is needed for object references in our JSON-LD).

### DIFF
--- a/docs/1.2-DRAFT/appendix/jsonld.md
+++ b/docs/1.2-DRAFT/appendix/jsonld.md
@@ -97,7 +97,7 @@ The order of the `@graph` array is not significant. Above we see that the RO-Cra
 
 ## Describing entities in JSON-LD
 
-Properties of an entity can refer to another URL or entity by using the form `{"@id": "uri-reference"}` as in the example above, where the [author] property in the [File] entity refer to the [Person] entity, identified as `#alice`. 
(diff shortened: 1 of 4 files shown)


stain:
Rendered at https://stain.github.io/ro-crate/1.2-DRAFT/appendix/jsonld#add-local-definitions-of-ad-hoc-terms

stain:

PR #233, opened 2023-01-17
Ad-hoc terms can be moved to a Profile Crate

stain:
Links the [Ad-hoc section](https://www.researchobject.org/ro-crate/1.2-DRAFT/appendix/jsonld.html#adding-new-or-ad-hoc-vocabulary-terms) to the new [profile section](https://www.researchobject.org/ro-crate/1.2-DRAFT/profiles) and suggest using that profile as a `@context` to avoid repeating terms.

PR #231, opened 2023-01-16
Indicate encoding format profile using conformsTo

stain:
Example: https://by-covid.github.io/baseline-use-case-synthetic-crate/ro-crate-preview.html#vaccines_effectiveness_common_data_model_specification/data/metadata/attributes.csv

```json
 {
  "@id": "attributes.csv",
  "@type": "File",
  "encodingFormat": ["text/csv", {"@id": "https://www.nationalarchives.gov.uk/PRONOM/x-fmt/18"}],
  "conformsTo": {"@id": "https://docs.ropensci.org/dataspice/#create-spice"}
 },
 {
  "@id": "https://www.nationalarchives.gov.uk/PRONOM/x-fmt/18",
  "@type": "WebPage",
  "name": "Comma-Separated Values format (CSV)"
 },
 {
  "@id": "https://docs.ropensci.org/dataspice/#create-spice",
  "@type": "CreativeWork",
  "name": "dataspice CSV profile"
 }
```

This also adds note to explain conformsTo used on ro-crate-metadata.json and root dataset for profiles (assumes #229)

PR #235, opened 2023-01-25
Use W3C Profile vocabulary as part of Profile Crate

stain:
Expand [profile crate](https://www.researchobject.org/ro-crate/1.2-DRAFT/profiles#profile-crate) to formalize the elements of the profile using the [W3C Profile vocabulary](https://www.w3.org/TR/dx-prof/)


stain:
See https://stain.github.io/ro-crate/1.2-DRAFT/profiles#profile-crate using type `Profile`, and use of `hasArtifact` to a `ResourceDescriptor` in https://stain.github.io/ro-crate/1.2-DRAFT/profiles#what-is-included-in-the-profile-crate

(Rendered together with #233 #232 #231 #230 #229)

Issue #251, opened 2023-05-02
Issue: A number of new @context terms in 1.2 that don't resolve to human readble definitions 


ptsefton:
Not sure what we can do about this but a number of new terms have been added that are not from Schema.org

Eg:
```

          "hasArtifact": "http://www.w3.org/ns/dx/prof/hasArtifact",
          "hasResource": "http://www.w3.org/ns/dx/prof/hasResource",
          "hasRole": "http://www.w3.org/ns/dx/prof/hasRole",
          "hasToken": "http://www.w3.org/ns/dx/prof/hasToken",
          "isInheritedFrom": "http://www.w3.org/ns/dx/prof/isInheritedFrom",
          "isProfileOf": "http://www.w3.org/ns/dx/prof/isProfileOf",
          "isTransitiveProfileOf": "http://www.w3.org/ns/dx/prof/isTransitiveProfileOf",
          "ResourceDescriptor": "http://www.w3.org/ns/dx/prof/ResourceDescriptor",
          "ResourceRole": "http://www.w3.org/ns/dx/prof/ResourceRole",
          "Profile": "http://www.w3.org/ns/dx/prof/Profile",
```

These do not resolve to HTML pages and some of them look like they may cause confusion as they are similar to existing terms eg `hasResource`  vs `hasPart` and `hasRole` vs Schema.org's `Role` class. If we need these terms we may have to put up some pages that provide descriptions and add a mechanism for Crate viewer applications to link to the online definitions.

stain:
Agree that this is not pretty, w3 has done better job on others like https://www.w3.org/ns/prov -- we're meeting the Profile editor later this month so can also request a HTML.

Should we anyway add these to the eat-own-dogfood RO-Crate https://www.researchobject.org/ro-crate/1.1/ro-crate-preview.html (but fot 1.2) using `DefinedTerm`?  Then it can become the base Profile Crate.

The terms are also defined in https://www.researchobject.org/ro-crate/1.2-DRAFT/metadata.html#additional-metadata-standards (which would not be machine readable) but perhaps that should also link to the corresponding HTML definition.

stain:
Previewer adds (?) link per term and these term don't resolve to anything useful for people.

Issue #160, opened 2021-07-08
Use Case: Resolve RO-Crate from persistent identifier/DOI with landing page


stain:
As a **type of user**, I want **some goal** so that **some reason**.

As a **potential programmatic consumer** of RO-Crates I want to **find the RO-Crate metadata file** given a persistent identifier/DOI so that I can **index/catalogue potential Crates**.

As discussed in [RO-Crate call 2021-07-08](https://s.apache.org/ro-crate-minutes#heading=h.6w70pxl7gwvd) and already explored in #154 this should be with a couple of options:

A persistent identifier for a creative work SHOULD for browsers give a human-readable representation like HTML. This MAY be equivalent to the [RO-Crate Website](https://www.researchobject.org/ro-crate/1.1/structure.html#ro-crate-website-ro-crate-previewhtml-and-ro-crate-preview_files), or a more specific rendering that just happens to have a corresponding RO-Crate.

To resolve a persistent identifier to a machine-readable JSON-LD, these approaches are recommended to retrieve its [RO-Crate metadata file](root-data-entity.md#ro-crate-metadata-file-descriptor):

1. [HTTP Content-negotiation](https://httpd.apache.org/docs/2.4/content-negotiation.html) for the [RO-Crate media type](appendix/jsonld.md#ro-crate-json-ld-media-type), for example:  
  Requesting `https://w3id.org/ro/profile/paradisec/0.1` with HTTP header  
  `Accept: application/ld+json;profile=https://w3id.org/ro/crate` redirects to the _RO-Crate Metadata file_
  `https://example.org/ro-profiles/paradisec-0.1.0/ro-crate-metadata.json`
2. [FAIR Signposting](https://signposting.org/FAIR/) in HTTP headers response to `HEAD` request, using `rel="describedby"` and the [RO-Crate media type](appendix/jsonld.md#ro-crate-json-ld-media-type):
   `Link: <https://example.org/workflows/29/ro-crate-metadata.json>; rel="describedby"; type="application/ld+json;profile=https://w3id.org/ro/crate"`
3. Parse the Landing Page HTML, looking for FAIR Signposting `<link href="…" rel="describedby" type="…">`  as above, or `<script type="application/ld+json">` blocks embedding the metadata file as for [RO-Crate Website](https://www.researchobject.org/ro-crate/1.1/structure.html#ro-crate-website-ro-crate-previewhtml-and-ro-crate-preview_files). (Note: The `<script>` type do not include `profile`)
[...]

PR #255, opened 2023-05-11
Specify how cite-as and Signposting should be used

### LINKED ISSUE #228: Use Case: How to get the contents of a RO as a zip file?
As a programmer, I want to obtain the aggregated contents of a Research Object as a downloadable resource. 

Ideally, I would like to do so through a request and content-negotiation. But I do not see an agreement about how to serve the RO-Crate itself. Can we agree into something like `application/zip`?
Can we have some community-agreed guidelines?

### LINKED ISSUE #239: BUG: Error in @context  cite-as does not resolve
**Describe the bug**
 The default context has a definition of cite-as which does not resolve. https://www.w3.org/ns/iana/link-relations/relation#cite-as

(I want to use this property to put a text citation on a crate so it can say "Cite this crate as ...")

**URL**
https://www.researchobject.org/ro-crate/1.1/context.jsonld

stain:
This fixes #239 as re-opened by @ptsefton to specify `cite-as` as well as `identifier` in the case of a persistent identifier.

I also added the profile definition links as part of #251. I think some of these can be removed as they are not referenced in spec.

The Signposting bit should later be consolidated with the "How to retrieve a crate" currently in `profiles.md` to fix #228 #160.

Catering Peter's question on textual citation I added `description` on the `PropertyValue` -- not sure what you think about that.

### DIFF
--- a/docs/1.2-DRAFT/metadata.md
+++ b/docs/1.2-DRAFT/metadata.md
@@ -102,23 +102,26 @@ RO-Crate also uses the _Portland Common Data Model_ ([PCDM] version <https://pcd
 {: .note }
 > The terms `RepositoryObject` and `RepositoryCollection` are renamed to avoid collision between other vocabularies and the PCDM terms `Collection` and `Object`. The term `RepositoryFile` is renamed to avoid clash with RO-Crate's `File` mapping to <http://schema.org/MediaObject>.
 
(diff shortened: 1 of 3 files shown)


stain:
@ptsefton do you think this is OK to add, or shall we delay it past 1.2? Can split it so the `[definition]` links can be added only, but I heard you wanted to add `cite-as` as well

ptsefton (approved):
This looks fine

ptsefton (approved):

PR #256, opened 2023-05-11
Add DefinedTerm for Profiles vocabulary, PCDM and cite-as

### LINKED ISSUE #251: Issue: A number of new @context terms in 1.2 that don't resolve to human readble definitions

stain:
Together with #255 this fixes #251.

Note that I didn't define the terms with rdfs as they are already defined such upstream. For PCDM the `termCode` reflects the term used in the JSON-LD context as specified in https://www.researchobject.org/ro-crate/1.2-DRAFT/profiles#extension-terms

I added `inDefinedTermSet` back reference to the vocabulary but didn't list all their members, as the remaining classes/properties are NOT part of the RO-Crate core profile.

### DIFF
--- a/docs/1.2-DRAFT/metadata.md
+++ b/docs/1.2-DRAFT/metadata.md
@@ -110,10 +110,6 @@ RO-Crate use the [Profiles Vocabulary](https://www.w3.org/TR/2019/NOTE-dx-prof-2
 - `hasArtifact` mapped to <http://www.w3.org/ns/dx/prof/hasArtifact>
 - `hasResource` mapped to <http://www.w3.org/ns/dx/prof/hasResource>
 - `hasRole` mapped to <http://www.w3.org/ns/dx/prof/hasRole>
-- `hasToken` mapped to <http://www.w3.org/ns/dx/prof/hasToken>
-- `isInheritedFrom` mapped to <http://www.w3.org/ns/dx/prof/isInheritedFrom>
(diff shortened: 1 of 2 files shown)


ptsefton (approved):

ptsefton:
Ok this looks like a good approach

PR #257, opened 2023-05-12
Add RepositoryFile to the context

simleo:
https://www.researchobject.org/ro-crate/1.2-DRAFT/metadata#additional-metadata-standards says RO-Crate imports `RepositoryFile` among other PCDM terms, but the term is not actually in the context.


stain (approved):
Added also in schema-context.py

Could arguably be added in 1.1 as well, but I don't think we'll do another point release

PR #259, opened 2023-05-17
contentUrl for direct download, add section on Converting to Attached RO-Crate

stain:
New section _Converting from Detached to Attached Crate_ with a suggested algorithm and warnings it may be elaborate. It notes that if the crate is an Attached Crate on the Web then it is easier due to relative paths.

The "Files that are also on the web" section adds `contentUrl` for the purpose of direct download, while `url` can be used for landing pages only - as proposed in https://github.com/ResearchObject/ro-crate/pull/189#issuecomment-1227697348

This completes #189 as this section was marked as `TODO`.

Tagging @jmfernandez @ptsefton @simleo which had views on retrieval in #189.

### DIFF
--- a/docs/1.2-DRAFT/appendix/relative-uris.md
+++ b/docs/1.2-DRAFT/appendix/relative-uris.md
@@ -126,7 +126,40 @@ If the new Detached RO-Crate is not meant as a snapshot of the corresponding Att
 
 ## Converting from Detached to Attached RO-Crate
 
-_TODO_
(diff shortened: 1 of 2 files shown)


ptsefton:
I thinnk it's worth revisiting this to see if there is a way to specify where files should be saved when converting from detatched to attached crate -- a "saveAs" relative URI equivalent to @id in an attached crate. We discussed this a long time ago and didn't reach a conclusion.

ptsefton:
This could be done with an identifier->PropertyValue that specifies the local path

PR #270, opened 2023-07-14
Profiles for ISA and ARC

floWetzels:
This PR updates the ARC section in `profiles.md` and adds the ISA profile (the main part of an ARC).
The ISA profile now contains an explicit description of the investigation profile and is referenced by the ARC profile.
Further references and descriptions in the ARC profile were updated/corrected.


stain (approved):

stain:
Thanks, @floWetzels! Should soon be live

PR #272, opened 2023-08-02
Add Sciebo-RDS to Tools

stain:
Probably should write up an in-use story as well
https://www.research-data-services.org/


stain:
https://www.research-data-services.org/documentation/development/contributing/developing-connectors/ passes RO-Crate as part of the connector API

Issue #275, opened 2023-09-29
Add CodeMeta terms


stain:
As agreed in RO-Crate call 2023-09-28, we will add the additional [CodeMeta terms](https://codemeta.github.io/terms/) to the RO-Crate context.

Some considerations before we merge this:

* CodeMeta 3.0 from July is not quite released yet -- see https://github.com/codemeta/codemeta/issues/236#issuecomment-1740196595
* @dgarijo has established https://w3id.org/codemeta and https://w3id.org/codemeta/3.0 as PIDs for their context (see codemeta/codemeta#321) 
  -  as usual we'll anyway avoid nested context resolution and have our own fixed declarations directly in our context
* Their namespace https://codemeta.github.io/terms/ is not persistent as it relies on `github.io`  (codemeta/codemeta#216)
* Terms like https://codemeta.github.io/terms/readme  are 404 Not Found, which we don't really want RO-Crate users to deal with.
  - This is not helped by there not being any ontology or vocabulary beyond their JSON-LD context. We don't need that, but at least the link should go to something.
  - @dgarijo has proposed using <https://w3id.org/codemeta#> as namespace, which I support. But this may mean waiting for a 4.0 release. Possibly a 3.1 can be done quicker, as 3.0 is not publicly announced as far as I can see.
* Schema.org terms `maintainer` and `funding` are redefined by Codemeta (see codemeta/codemeta#274 and codemeta/codemeta#232)
  - I've commented out these two, as our [context-generator](https://github.com/ResearchObject/ro-crate/blob/master/scripts/schema-context.py) luckily bails out on conflicts

Given these uncertainties I am reluctant to add this directly, but I've prepared a branch we can keep ready.  I've commented (perhaps too much!) in these issues to perhaps help move things forward.

stain:
Ping @CaroleGoble who wanted a list of issues!

elichad:
Fixed by #276

elichad:
Good catch @simleo - looks like #276 only added the terms to the 'compatibility' version of `context.jsonld` (meaning they show up in https://www.researchobject.org/ro-crate/1.2-DRAFT/context.jsonld, but not the new link you shared which should be the main link). Reopening so we can make sure this gets fixed.

PR #276, opened 2023-09-29
Add Codemeta terms

stain:
See namespace considerations in #275 before merging.


stain:
Not sure if we should add this for 1.2 when we don't have a corresponding section explaining when to use `embargoEndDate` etc.  Suggest to delay this for 2.0.

docs/1.2-DRAFT/context.jsonld
```diff
@@ -2670,7 +2673,7 @@
           "starRating": "http://schema.org/starRating",
           "startDate": "http://schema.org/startDate",
           "startOffset": "http://schema.org/startOffset",
-          "startTime": "http://schema.org/startTime",
+          "startTime": "http://schema.org/startTime",embargoEndDate
```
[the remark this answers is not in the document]
jmfernandez: After having a look at the context, I also think it is a typo

stain:
@dgarijo  to double-check the PIDs

docs/1.2-DRAFT/context.jsonld
```diff
@@ -2892,6 +2895,16 @@
...
+          "softwareSuggestions": "https://w3id.org/codemeta#softwareSuggestions",
+          "continuousIntegration": "https://w3id.org/codemeta#continuousIntegration",
+          "buildInstructions": "https://w3id.org/codemeta#buildInstructions",
+          "developmentStatus": "https://w3id.org/codemeta#developmentStatus",
+          "embargoEndDate": "https://w3id.org/codemeta#embargoEndDate",
+          "readme": "https://w3id.org/codemeta#readme",
+          "issueTracker": "https://w3id.org/codemeta#issueTracker",
+          "referencePublication": "https://w3id.org/codemeta#referencePublication",
+          "hasSourceCode": "https://w3id.org/codemeta#hasSourceCode",
+          "isSourceCodeOf": "https://w3id.org/codemeta#isSourceCodeOf",
```
dgarijo, line 2907: As of today, the identifier is https://codemeta.github.io/terms/

I have been pushing to shift it to https://w3id.org/codemeta/, which is where we currently have the contexts hosted. But it's not yet the case.

docs/1.2-DRAFT/metadata.md
```diff
@@ -151,6 +151,22 @@ To support geometry in [Places](contextual-entities.md#places), these terms from
...
+* `buildInstructions` mapped to <https://w3id.org/codemeta#buildInstructions>
+* `developmentStatus` mapped to <https://w3id.org/codemeta#developmentStatus>
+* `continuousIntegration` mapped to <https://w3id.org/codemeta#continuousIntegration>
+* `embargoEndDate` mapped to <https://w3id.org/codemeta#embargoEndDate>
+* `hasSourceCode` mapped to <https://w3id.org/codemeta#hasSourceCode>
+* `isSourceCodeOf` mapped to <https://w3id.org/codemeta#isSourceCodeOf>
+* `issueTracker` mapped to <https://w3id.org/codemeta#issueTracker>
+* `readme` mapped to <https://w3id.org/codemeta#readme>
+* `referencePublication` mapped to <https://w3id.org/codemeta#referencePublication>
+* `softwareSuggestions` mapped to <https://w3id.org/codemeta#softwareSuggestions>
```
dgarijo, line 165: Same as above. The current namespace is https://codemeta.github.io/terms/, even if that one is not machine-readable

scripts/schema-context.py
```diff
@@ -166,6 +175,25 @@ def main():
...
+          # https://github.com/codemeta/codemeta/issues/160
+          #("funding", "https://w3id.org/codemeta#funding"),
+          ("readme", "https://w3id.org/codemeta#readme"),
+          ("issueTracker", "https://w3id.org/codemeta#issueTracker"),
+          ("referencePublication", "https://w3id.org/codemeta#referencePublication"),
+          # Maintainer is already in schema.org, see
+          # https://github.com/codemeta/codemeta/pull/250
+          #("maintainer", "https://w3id.org/codemeta#maintainer"),
+          ("hasSourceCode", "https://w3id.org/codemeta#hasSourceCode"),
+          ("isSourceCodeOf", "https://w3id.org/codemeta#isSourceCodeOf"),
```
dgarijo, line 196: Same issue as above. Current correct ns is  https://codemeta.github.io/terms/

dgarijo:
I added a few comments @stain. In Codemeta we resolved the resolution and content negotiation of w3id.org/codemeta, but at this time, the id is still ` https://codemeta.github.io/terms/`. I followed with another comment to see if we can push forward changing the id of the vocabulary. That may lead to incompatibility changes, so I am not sure how reluctant will people be.

I suggest using for now the current Codemeta id, even if it's not machine readable

stain:
It's OK, better remain compatible for now and use codemeta.github.io namespace.

I can add the machine readable definitions to our core [Profile Crate](https://github.com/ResearchObject/ro-crate/blob/master/docs/1.2-DRAFT/ro-crate-metadata.json#L389) for 1.2-DRAFT, but we would need to add also human-readable text to explain also why these URLs are 404 Not Found.

dgarijo:
As shown in https://github.com/codemeta/codemeta/discussions/360 my proposal is to move it to https://w3id.org/codemeta/terms/

It would be great if you drop a +1 to show your support to this change

dgarijo:
Hopefully by June 20th we'll get it solved. But community standards are slow sometimes.

stain:
I've added machine readable version in https://github.com/ResearchObject/ro-crate/blob/codemeta/docs/1.2-DRAFT/ro-crate-metadata.json#L736 , which you are free to adapt/correct. Unsure about if `domainIncludes` should always be `SoftwareApplication`

dgarijo:
I will review when time allows. Thanks!

stain:
Ideally from https://www.researchobject.org/ro-crate/specification/1.2-DRAFT/appendix/jsonld.html#extending-ro-crate we should have a `sameAs` link for these term so that RO-Crate previewer has something to link to instead of the 404 Not Found pages https://codemeta.github.io/terms/developmentStatus 

However, while each term is defined on https://codemeta.github.io/terms/ textually, there are no HTML anchors there, so https://codemeta.github.io/terms/#developmentStatus for instance would not help much and would still have that dangerous `github.io` in its non-persistent URL:.  

And we can't have multiple `DefinedTerm` entities  with the same `"sameAs": "https://codemeta.github.io/terms/"`  as that would semantically link them all together to the same term.

@dgarijo  Could we use the `https://w3id.org/codemeta#` prefix in `sameAs`? 

```json
   {
    "@id": "https://codemeta.github.io/terms/developmentStatus",
    "@type": ["DefinedTerm", "rdf:Property"],
    "name": "developmentStatus",
    "termCode": "developmentStatus",
    "description": "Description of development status, e.g. Active, inactive, suspended. See repostatus.org",
    "domainIncludes": {"@id": "http://schema.org/SoftwareSourceCode"},
    "rangeIncludes": {"@id": "http://schema.org/Text"}
    "sameAs": "https://w3id.org/codemeta#developmentStatus"
    }

```

While https://w3id.org/codemeta#developmentStatus does resolve to a HTML page you still have to scroll down and look for "developmentStatus" but at least it's a persistent URL for each term, and also in the promised namespace-to-come.

Related question is if `sameAs` should rather be as an object reference  to match with schema.org JSON-LD mapping as an object property:

```json
"sameAs": { "@id": "https://w3id.org/codemeta#developmentStatus" }
```

stain:
@dgarijo  could you also check if I got the `domainIncludes` and `rangeIncludes` correct? https://github.com/ResearchObject/ro-crate/pull/276/files#diff-391353380a9c2b594e12f6735db992305846175ab2c59389fc314ee9c5aafa37R767

stain:
I got it wrong above, I see https://github.com/codemeta/codemeta/discussions/360#discussioncomment-9519453 suggests https://w3id.org/codemeta/terms/ as namespace, so I've added that, e.g. 

```json
{
    "@id": "https://codemeta.github.io/terms/buildInstructions",
    "@type": ["DefinedTerm", "rdf:Property"],
    "sameAs": "https://w3id.org/codemeta/terms/buildInstructions"
}

stain:
@dgarijo Can you confirm I got this right now..?

dgarijo:
Will complete a review by the end of the week.

dgarijo (approved):
Thanks for updating this. It looks good.
I think some of the domains may be also SoftwareApplication, but let us leave it consistent with https://github.com/codemeta/codemeta/blob/master/crosswalk.csv#L63.
Please note that developmentStatus is under discussion, the range may change to a URL. But I don't know when.
Nice idea adding the sameAs to the `/terms` ns. I will try and resurrect the thread once September starts.

PR #278, opened 2023-10-06
Tightening up wording around data entities -- we had not actual said …

### LINKED ISSUE #277: Bug: We are not explicit enough about Data Entities -- MUST or SHOULD they exist in root directory?
**Describe the bug**

Working with a project where a developer has created a number of crates with File entities that don't exist on disk, and included a number of files that are not described in the Crate in any way, I realised that we are not explicit enough about expectations for whether files should be present in a crate.

**URL**
https://github.com/ResearchObject/ro-crate/blob/master/docs/1.2-DRAFT/data-entities.md

**Suggested fix**

I will do a PR with some enhanced wording. This is intended to:

- Make it clear that while you don't have to describe all files, you SHOULD explain what's in directories of un-described files and why they are not described individually.
- Make it clear that it is at leat not recommended to describe files that don't exist (or should we make this a MUST NOT?)

**Additional context**

ptsefton:
…whether Data Entities MUST/SHOULD exist - fixes #277


docs/1.2-DRAFT/data-entities.md
```diff
@@ -214,6 +220,11 @@ Additionally, `Dataset` entities SHOULD have:
 
 Any of the properties of schema.org [Dataset] MAY additionally be used (adding contextual entities as needed). [Directories on the web](#directories-on-the-web-dataset-distributions) SHOULD also provide `distribution`.
 
+## Local Data Entities
+
+Where an RO-Crate is used to describe _files_  and _directories_ contained within the _RO-Crate root_ directory these locally referenced _Data Entities_ as described in the _RO-Crate Metadata Document_  SHOULD be present in the _RO-Crate root_.
```
ptsefton, line 225: Should we say MUST here? I think the python library interprets this as a MUST -- we have not made our intent clear in the spec...

docs/1.2-DRAFT/data-entities.md
```diff
@@ -174,7 +180,7 @@ If there is no web-accessible description for a file format it SHOULD be describ
 
 ## Core Metadata for Data Entities
 
-The table below outlines the properties that Data Entities, when present, MUST have to be minimally valid.
+The list below outlines the properties that Data Entities, when present in the _RO-Crate Metadata Document_, SHOULD exist, either in `RO-Crate Root` directory or on the web.
```
simleo, line 183: It looks like two different sentences got mixed up in the edit. Also, there is no listing or table below, and not all recommendations are `MUST`. I suggest changing the paragraph to:

Below is an outline of properties that Data Entities, when present in the _RO-Crate Metadata Document_, MUST or SHOULD have. Such data entities should exist, either in the RO-Crate Root directory or on the web.

docs/1.2-DRAFT/data-entities.md
simleo: No, the Python library allows data entities to be missing from the directory. This was requested by you in https://github.com/ResearchObject/ro-crate-py/issues/73. See also https://github.com/ResearchObject/ro-crate-py/issues/135, where changes are made so that such crates are "not the norm".

ptsefton: Ok, yes, software libraries need to be able to read things that are not fully valid, so they can report warnings and errors -- the point I was making that the python implementers assumed that if you reference a file it SHOULD/MUST be there (which is perfectly reasonable) - the original point of the spec was to describe files -- it doesn't make sense to be shipping crates that contain references to things that aren't there.

ptsefton: For  RO-Crate version 2 we may want to formally define what constitutes and error or a warning.

docs/1.2-DRAFT/data-entities.md
ptsefton: There is no table - but there is a list, I was just trying to reflect that.

PR #281, opened 2023-12-15
Add Autosubmit use case with climate workflows and HPC using RO-Crate.

kinow:


kinow:
I think the `in-use/index.md` format changed. Updated PR to match it, and created a new page under the RO-Crate in Use collection (I believe that makes the new item magically appear in the page).

stain (approved):

stain:
Thanks, could we try to add it also to https://github.com/researchobject/ro-crate-2024 ? It needs a couple extra keys in the markdown metadata fields there though.

kinow:
> Thanks, could we try to add it also to https://github.com/researchobject/ro-crate-2024 ? It needs a couple extra keys in the markdown metadata fields there though.

Sure thing, will send a PR there today. Thanks Stian!

PR #290, opened 2024-02-27
Updated ARC Profile

floWetzels:
This PR updates the ARC and ISA profiles such that they now also specify the ISA types further down in the hierarchy like Assays, Studies or Protocols. They are based on the new/adapted bioschemas types LabProcess and LabProtocol. Previously, the profile has only been specified for the top level Investigation object.


stain (approved):

stain:
Thanks, @floWetzels  ! Should soon be live.

PR #291, opened 2024-02-27
Add geosparql vocabulary terms

### LINKED ISSUE #286: Add geo terms to context
These need to be in the context (and defined so they can be referenced in text like [Geometry])

```json
 {
      "Geometry": "http://www.opengis.net/ont/geosparql#Geometry",
      "asWKT": "http://www.opengis.net/ont/geosparql#asWKT"
    }

```

stain:
This fixes #286.

Note that only two GeoSPARQL terms are included. The rest can be accessed through prefix "geosparql", we won't use prefix "geo" as it would conflict with https://schema.org/geo which will be used to point to the `Geometry` object.

See https://opengeospatial.github.io/ogc-geosparql/geosparql11/spec.html for more information.

### DIFF
--- a/docs/1.2-DRAFT/context.jsonld
+++ b/docs/1.2-DRAFT/context.jsonld
@@ -2892,6 +2892,8 @@
           "ResourceDescriptor": "http://www.w3.org/ns/dx/prof/ResourceDescriptor",
           "ResourceRole": "http://www.w3.org/ns/dx/prof/ResourceRole",
           "Profile": "http://www.w3.org/ns/dx/prof/Profile",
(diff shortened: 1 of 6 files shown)


ptsefton (approved):
Looks right to me

simleo:
These URLs all lead to 404 / not found errors:

http://www.opengis.net/ont/geosparql#Geometry
http://www.opengis.net/ont/geosparql#asWKT
https://opengeospatial.github.io/ogc-geosparql/geosparql11/#Geometry
https://opengeospatial.github.io/ogc-geosparql/geosparql11/#asWKT
https://opengeospatial.github.io/ogc-geosparql/geosparql11/spec.html

PR #292, opened 2024-03-12
add example for directories on the web to specification

dnlbauer:
The documentation talks about how directories on the web that are not distributed as archives should be included. The docu also recommends to include all files inside such a directory as `hasPart`. However, no example is given for this. The only example shows how archives can be linked.

This PR adds an example that links to a web dir that also lists individual files of the dataset.


stain:
Thanks, @dnlbauer! Merged. 

Would you join https://www.researchobject.org/ro-crate/community.html (issue #1) to get attribution?

PR #295, opened 2024-04-12
Fix unreadable SVG

### LINKED ISSUE #294: Unreadable SVG in introduction documentation
**Describe the bug**
SVG in the introduction section seems to show an image and additional text, making it unviewable.

image

**URL**
Documentation page: https://www.researchobject.org/ro-crate/1.2-DRAFT/introduction.html
Direct link to the SVG: https://www.researchobject.org/ro-crate/assets/img/crate1-folders.svg

**Suggested fix**
Place the text beneath the image layer or remove it from the SVG? (I'm not sure if aria-label or the text is preferred here if it was included for accessibility.)

**Additional context**
I also dug through the SVG HTML and it looks like it was exported from Inkscape (💖) with all of the Inkscape metadata. It likely doesn't matter for an SVG this small, but it can be helpful to export as an Optimized SVG for file size, which will strip out some of the extra data.

elichad:
Should fix #294.

I am not sure that the text/aria-labels within the SVG are even exposed to screen readers when we import the file using `` - when I checked with NVDA/Firefox it only read the content in `alt`. But I followed pattern 11 from Deque's [Creating Accessible SVGs](https://www.deque.com/blog/creating-accessible-svgs/) anyway, as it may be different for other screen reader/browser combinations.

### DIFF
--- a/docs/assets/img/crate1-folders.svg
+++ b/docs/assets/img/crate1-folders.svg
@@ -7,7 +7,7 @@
    viewBox="0 0 168.80416 66.939585"
    version="1.1"
    id="svg5"
-   aria-labelledby="lightbulb11 description11"
+   aria-labelledby="title_crate1 description_crate1"
    inkscape:version="1.3 (0e150ed6c4, 2023-07-21)"
    sodipodi:docname="crate1-folders.svg"
    inkscape:export-filename="crate1-folders.svg"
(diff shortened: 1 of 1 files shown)


stain:
File also in:

```
(base) stain@xena:~/src$ find . -name 'crate1-folders.svg' 
./ro-crate/docs/assets/img/crate1-folders.svg
./galaxy-training-material/topics/fair/images/ro-crate-intro/crate1-folders.svg
./galaxy-training-material/_site/training-material/topics/fair/images/ro-crate-intro/crate1-folders.svg
./packaging_data_with_ro-crate/fig/crate1-folders.svg
./ro-crate-2024/assets/img/crate1-folders.svg
```

PR #296, opened 2024-04-25
How to reference and retrieve another RO-Crate

### LINKED ISSUE #228: Use Case: How to get the contents of a RO as a zip file?

stain:
This PR fixes #228 #160

Generalizes the Content-negotiate-or-signposting section from not just Profile Crates.

For ZIP files this is still vague in that it says _If the retrieved resource is a ZIP file (`Content-Type: application/zip`), then extract `ro-crate-metadata.json`, or, if the archive root only contains a single folder (e.g. `folder1/`), extract `folder1/ro-crate-metadata.json`_  
I've also added BagIt reference as this would be a second folder, e.g. `folder1/data/ro-crate-metadata.json` and then the checksums should be verified first as we do in https://trefx.uk/5s-crate/0.4/#check-phase

As for referencing another RO-rate from another, either the referenced RO-Crate can have its own `distribution` with a `conformsTo`:

```json
  {
    "@id": "./",
    "@type": "Dataset",
    "identifier": "https://doi.org/10.48546/workflowhub.workflow.775.1",
    "url": "https://workflowhub.eu/workflows/775/ro_crate?version=1",
    "name": "Research Object Crate for Jupyter Notebook Molecular Structure Checking",
    "distribution": {"@id": "https://workflowhub.eu/workflows/775/ro_crate?version=1"},
    "…": ""
  },
  {
    "@id": "https://workflowhub.eu/workflows/775/ro_crate?version=1",
    "@type": "DataDownload",
    "encodingFormat": ["application/zip", {"@id": "https://www.nationalarchives.gov.uk/PRONOM/x-fmt/263"}],
    "conformsTo": { "@id": "https://w3id.org/ro/crate" }
  }
```

or it can have a `subjectOf` to a `ro-crate-metadata.json`:

```json
{
  "@id": "http://example.com/another-crate/",
  "@type": "Dataset",
  "conformsTo": { "@id": "https://w3id.org/ro/crate" },
  "subjectOf": { "@id": "http://example.com/another-crate/ro-crate-metadata.json" }
},
{
  "@id": "http://example.com/another-crate/ro-crate-metadata.json",
  "@type": "CreativeWork",
  "encodingFormat": "application/ld+json"
}
```

As used by the 5s-crate profile: https://trefx.uk/5s-crate/0.4/#referencing-a-workflow-crate


stain:
Could  @dgarijo or @ptsefton  have a look at this? I've used it here: 
https://stain.github.io/workflow-run-crate/profiles/0.5-DRAFT/process_run_crate/ro-crate-preview.html#https%3A//www.researchobject.org/workflow-run-crate-paper/mapping/

Perhaps we should add that `isPartOf` pattern as well on how to mention a file within another crate? (Could get tricky to make absolute URIs..)

dgarijo:
Thanks @stain , I have had a look. The only thing that it is not fully clear to me is where the distribution information is supposed to be added: is it on the crate referencing the other crate, or in the referenced crate metadata?

For example, let's say crate A references crate B. Usually I would add a link in A to B. But here you recommend adding also where B is stored, correct? As opposed to adding a link to B, and hoping that when I resolve the id I get a JSON-LD with the distribution information.

The only potential issue I see is that distributions may not have persistent ids. If the link from A to B persists, but the distribution is hosted elsewhere in the meantime, B has no means to tell A about this. But I am ok with this limitation

docs/1.2-DRAFT/data-entities.md
```diff
@@ -389,7 +389,52 @@ These can be included for File Data Entities as additional metadata, regardless
 
 ### Directories on the web; dataset distributions
 
-A _Directory File Entry_ or [Dataset] identifier expressed as an absolute URL on the web can be harder to download than a [File] because it consists of multiple resources. It is RECOMMENDED that such directories have a complete listing of their content in [hasPart], enabling download traversal.
+A _Directory File Entry_ or [Dataset] identifier expressed as an absolute URL on the web can be harder to download than a [File] because it consists of multiple resources. It is RECOMMENDED that such directories have a complete listing of their content in [hasPart], enabling download traversal, or are themselves RO-Crates.
+
+#### Referencing other RO-Crates
+
+A referenced RO-Crate is also a [Dataset], but where its [hasPart] do not need to be listed. Instead, its content and further metadata is available from its own RO-Crate Metadata File:
+
```
dgarijo, line 397: Please clarify if this metadata goes in the source crate (crate referencing another crate) or the target crate (referenced crate).

dgarijo (changes requested):
Small clarification

stain:
Think this is good to go in now, waiting final approval from @dgarijo or @jmfernandez

stain:
I can't seem to resolve the  requested changes from @dgarijo although I have addressed them in earlier commits.

dgarijo:
Sorry @stain I am not available right now for review. Will do as soon as I can.

stain:
After meeting 2024-08-22, revised to not require `subjectOf` declaration if the RO-Crate metadata file can be resolved. This meant adding a section on how the absolute URI of the RO-Crate can be determined if there is no `identifier`

I now feel this should be split out into a page separate from data-entities.md

dgarijo:
Once again, apologies. Will review by the end of the week.

docs/_specification/1.2-DRAFT/data-entities.md
```diff
@@ -396,7 +396,98 @@ Note that if a local file is intended to be packaged within an _Attached RO-Crat
...
+
+This section defines how a _referencing_ RO-Crate ("A") can declare data entities within A's RO-Crate Metadata Document, in order to indicate a _referenced_ RO-Crate ("B"). There are different options on how to find the identifier to assign the referenced crate in A, and how a consumer of A finding such a reference can find the corresponding RO-Crate Metadata Document for B.
+
+##### Referencing RO-Crates that have a persistent identifier
+
+If the referenced RO-Crate B has a `identifier` declared as B's [Root Data Entity identifier](root-data-entity#root-data-entity-identifier), then this is a _persistent identifier_ which SHOULD be used as the URI in the `@id` of the corresponding entity in RO-Crate A.  For instance, if crate B had declared the identifier `https://pid.example.com/another-crate/` then crate A can reference B as an entity:
```
dgarijo, line 409: a `identifier` -> an `identifier`

docs/_specification/1.2-DRAFT/data-entities.md
```diff
@@ -396,7 +396,98 @@ Note that if a local file is intended to be packaged within an _Attached RO-Crat
...
+
+{.warning }
+> It is NOT RECOMMENDED to declare the generic profile `https://w3id.org/ro/crate` on a referencing crate A's own [root data entity](root-data-entity.html#direct-properties-of-the-root-data-entity), see [metadata descriptor](root-data-entity.html#ro-crate-metadata-descriptor). 
+
+Consumers that find a reference to a `Dataset` with the generic RO-Crate profile indicated MAY attempt to resolve the persistent identifier, but SHOULD NOT assume that the `@id` directly resolves to an RO-Crate Metadata Document. See section [Retrieving an RO-Crate](#retrieving-an-ro-crate) below for the recommended algorithm. 
+
+If an `identifier` is not declared in a referenced RO-Crate B, but the determined absolute URI has [Signposting] declared for a `Link:` with `rel=cite-as`, then that link MAY be considered as an equivalent permalink for B.
+
+
+##### Determening entity identifier for a referenced RO-Crate
```
dgarijo, line 430: Determening -> Determining

docs/_specification/1.2-DRAFT/data-entities.md
```diff
@@ -396,7 +396,98 @@ Note that if a local file is intended to be packaged within an _Attached RO-Crat
...
+
+If the RO-Crate Metadata Document is not available as a web resource, but only within an archive (e.g. ZIP), then instead reference it as a [Downloadable dataset](#downloadable-dataset).
+
+##### Referencing another metadata document
+
+If a referenced RO-Crate Metadata Document is known at a given URI or path, but its corresponding RO-Crate identifier can't be determined as above (e.g. [Retrieving an RO-Crate](#retrieving-an-ro-crate) fails or requires heuristics), then an referenced metadata descriptor entity SHOULD be added. For instance, if `http://example.com/another-crate/ro-crate-metadata.json` resolves to an RO-Crate Metadata Document describing root `./`, but `http://example.com/another-crate/` always return a HTML page without [Signposting] to the metadata document, then `subjectOf` SHOULD be added to an explicit metadata descriptor entity, which has `encodingFormat` declared for JSON-LD:
```
dgarijo, line 443: then an referenced metadata descriptor ->  then a referenced metadata descriptor

dgarijo (changes requested):
Minor editorial changes. It looks very nice. The bit on content negotiation helps  the issue I opened about retrieving contents, since you may add a distribution where to retrieve the zipped contents now (issue https://github.com/ResearchObject/ro-crate/issues/228)

docs/_specification/1.2-DRAFT/data-entities.md
```diff
@@ -416,9 +507,62 @@ Alternatively, a common mechanism to provide downloads of a reasonably sized dir
...
+    "identifier": "https://doi.org/10.48546/workflowhub.workflow.775.1",
+    "name": "Research Object Crate for Jupyter Notebook Molecular Structure Checking",
+    "distribution": {"@id": "https://workflowhub.eu/workflows/775/ro_crate?version=1"},
+    "…": ""
+  },
+  {
+    "@id": "https://workflowhub.eu/workflows/775/ro_crate?version=1",
+    "@type": "DataDownload",
+    "encodingFormat": ["application/zip", {"@id": "https://www.nationalarchives.gov.uk/PRONOM/x-fmt/263"}],
+    "conformsTo": { "@id": "https://w3id.org/ro/crate" }
```
jmfernandez, line 525: I understand that this is needed to be correct describing downloadable RO-Crate. But, in this scenario, if the graph does not contain an `ro-crate-metadata.json` (or similar) node, or contains it without a proper `conformsTo`, shall we assume that metadata should appear at the RO-Crate fetchable from the downloadable content?

jmfernandez (approved):
Although heuristics are properly described, I'm realizing we are going to need in some moment code snippet examples in a couple of programming languages to demonstrate how to properly fetch an RO-Crate with negotiation.

docs/_specification/1.2-DRAFT/data-entities.md
stain: This is a cheeky triple-use of our `conformsTo` as we define what the profile mean on different files.. so here the implication is that if there is `https://w3id.org/ro/crate` as profile of a `DataDownload`  then yes, there should be a `ro-crate-metadata.json` inside it's archive, which presumably would have its own (versioned) `conformsTo` declared.   Or if it's a `Dataset` with the same `conformsTo` then the metadata file is found differently.

We also use the same profile as part of the JSON-LD content type https://www.researchobject.org/ro-crate/specification/1.2-DRAFT/appendix/jsonld.html#ro-crate-json-ld-media-type 

You could argue this should be several different profile URIs.. I think it's a bit overloading, but if you squeeze your eyes a bit it's all saying "an RO-Crate" just in different ways in.

But we won't try to list the content of the zip-file in the referring crate, unless we need to (which would bring in `arcp` etc), so that's a nice simplification that we moved to having the references on the `Dataset` rather than requiring mention of the second RO-Crate Metadata Document.

stain: https://signposting.org/adopters/#workflowhub Signposting also uses the same profile for the ZIP file. But that should also be valid to mean "There will be an `ro-crate-metadata.json` inside, as stated in top of section "#### Retrieving an RO-Crate"

stain:
Thanks everyone, will merge now after fixing latest round of typos!  Then we see if we need to move it out.

PR #299, opened 2024-05-20
Made the language about hybrid attached/detatched crates stronger

ptsefton:
Following a discussion on Slack about how to describe files which are available locally AND on the web (in this case in ROHub) I have strengthened the wording to make it clear that in this case the `@id` should use a relative URI to point to the local file and contentUrl should point to the download.

Including the folks who were part of that conversation as reviewers.

### DIFF
--- a/docs/1.2-DRAFT/data-entities.md
+++ b/docs/1.2-DRAFT/data-entities.md
@@ -373,7 +373,7 @@ These MAY be included for File Data Entities as additional metadata, regardless 
 * [subjectOf] to a [CreativeWork] (or [WebPage]) that mentions this file or its content (but also other resources)
 * [mainEntityOfPage] to a [CreativeWork]  (or [WebPage]) that primarily describes this file (or its content) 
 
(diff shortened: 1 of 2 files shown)


docs/1.2-DRAFT/data-entities.md
```diff
@@ -378,15 +385,13 @@ These can be included for File Data Entities as additional metadata, regardless
     "subjectOf": {"@id": "http://example.com/reports/2019/annual-survey.html"}
   },
   {
-    "@id": "https://zenodo.org/record/3541888/files/ro-crate-1.0.0.pdf",
-    "@type": "File",
-    "name": "RO-Crate specification",
-    "encodingFormat": "application/pdf",
-    "identifier": "https://doi.org/10.5281/zenodo.3541888",
-    "url": "https://zenodo.org/record/3541888"
+    "@id": "http://example.com/reports/2019/annual-survey.html",
```
stain, line 388: Agree the `.pdf` should not be in here, but I added the entity for the `WebPage` landing page as referenced above.

stain (approved):

PR #301, opened 2024-05-28
Migrate to new website

elichad:
This is almost ready to go:
* needs the open question about use case text on https://github.com/ResearchObject/ro-crate-2024/pull/45 to be resolved
* GHA needs testing on my fork


elichad:
Migrated site now available at https://elichad.github.io/ro-crate/, this is ready for review

elichad:
Noting that some Makefile steps may now fail with this change - I haven't tested them and think it would be best to try/fix them after this PR is merged.

stain (approved):

PR #325, opened 2024-06-21
Updated profiles page with bugfixes in ISA profile

floWetzels:
Minor bugfixes in ISA profile: now uses correct schema.org terminology for the property `additionalType`.


stain:
Great, fixed so the hyperlink match your text.

PR #327, opened 2024-07-11
Updated LDaCA and related use cases and tools

ptsefton:
Mostly updates to things I'm involved in tho I did remove ROHub from the featured list as its RO-Crates are not  correct


stain:
Agreed in meeting 2024-07-11 that for now ROHub will be removed from "Featured" list due to some issues in its' RO-Crate representations.

dgarijo:
well, the RO-Crates are correct, but do not point to local resources, instead online resources right?
I could not make it to the meeting, but I think it's a shame we remove one of the portals that fully implement the spec.

rapw3k:
Tomorrow (friday 12th) we are releasing fix of the contextual entities to use recommended URI. We can discuss on chat

stain:
I would also prefer to keep the only portal for RO-Crates on the front page, perhaps we should check again after tomorrow's ROHub release.

@ptsefton  Could you take out the `featured` changes, then we can merge the rest of this PR now? There will be "issues" with pretty much any one we choose!

ptsefton:
I will redo as smaller PRs sorry everyone

rapw3k:
Hi, please check the latest changes in production I think most of the recommendations are now in place. The HTML preview will be improved in the next days

ptsefton:
@rapw3k - yes I did a new PR with just LDaCA and other Australian use cases in it. #329

I just downloaded the latest crate from rohub  as a zip https://www.rohub.org/c9df67a3-0d77-4029-8556-e62fcc95a35b?activetab=overview

The treatment of files has improved - 

image

Issue: Local paths are URL encoded `biblio%2F` should be `bibio/`

Question: There are some empty directories like tool -- I would have expected the notebook in here to be part of the download to get a complete research object? Only the output directory is populated. If that's the intent why are the empty directories included?



A couple of issues I found:

1. There is a context entry: https://w3id.org/ro/terms/earth-science# which resolves to a CSV file rather than JSON-LD (this causes the ro-crate-html-js renderer to crash. I'll put in an issue for that in ro-crate-js so it degrades more gracefully but you need to point that at a JSON file. (I removed that reference to get the preview to run)

2.  Contextual entiies are not linked with "@id" - they are string values
```
"creator": [
        "https://orcid.org/0000-0002-0206-7140",
        "mailto:jambyr@bas.ac.uk"
      ],
```

rapw3k:
hi @ptsefton 
* wrt folder encoding we can avoid it, we are checking
* wrt empty directories, this is normal and it should be like that. The resources inside the tools directory are external resources. External resources are not downloaded as part of the zip. You can see in manifest that these resources are part of the folder tool.
* empty directories are included because they exist physically in the research object, so they are part of the zip. 
* wrt context entry: https://w3id.org/ro/terms/earth-science# , i am not sure what tool you are using but the tool should support content negotiation and follow redirects. If you ask for json-ld you will get the json-ld (see below). This is in fact the recommended way of contributing to ro-crate terms (see https://www.researchobject.org/ro-terms/) 
curl -i -L -X GET \
   -H "Accept:application/ld+json" \
 'https://w3id.org/ro/terms/earth-science#'
* wrt contextual entities, we have made a new update today, please check it again. it should be fixed

rapw3k:
update:
* folder path encoding is fixed
* we also added the hasPart to folders (directories)

From my point of view now ro-crate is completely aligned with specification. If you find something else please let me know. This exercise was useful

PR #346, opened 2024-08-07
Update dataplant.md

stain:
See also https://f1000research.com/slides/13-776


elichad (approved):
Thanks for adding @stain - is it worth including the linked slides under Publications on this page as well?

PR #352, opened 2024-08-23
Add info about regional drop-ins

elichad:


docs/pages/community/community.md
```diff
@@ -117,7 +117,9 @@ The RO-Crate Community is open for anyone to [join us](https://github.com/Resear
 
 The RO-Crate team try to meet at least monthly in a telcon, see the [rolling agenda](https://s.apache.org/ro-crate-minutes) for schedule, call-in details and minutes.
 
-See also [recent and upcoming events](outreach#upcoming-and-recent-events).
+There are also regular [regional drop-in calls](https://docs.google.com/document/d/1l0EmDVE4k6oEZSoMGjes8mMNrJboRlmZW-kvjaeVqvE/edit#heading=h.twg8huxfni6c) aimed at those newer to the RO-Crate community. These sessions are suitable for beginner questions, as well as discussing ideas for new tools/integrations.
```
jmfernandez, line 120: Should we use a permalink, like in the RO-Crate minutes, for the Google Docs document holding regional drop-in calls minutes? So, if the document is moved or rotated in the future, the permalink does not have to change.

jmfernandez (commented):
Maybe a permalink pointing to the regional drop-in calls minutes should be created and used.

docs/pages/community/community.md
stain: added https://s.apache.org/ro-crate-regional

stain (approved):

PR #355, opened 2024-09-11
Fix minor errors in spec

### LINKED ISSUE #304: Issue in example of Workflow RO-Crate profile
Meanwhile I was carefully reading Workflow RO-Crate profile specification, I found that the example has an inconsistency. Next line

https://github.com/ResearchObject/ro-crate-2024/blob/42c15d25567d7e906ef43a5b53a20a397d500ab7/_specification/1.1/workflows.md?plain=1#L311

should be instead:

```json
      "@type": "ComputerLanguage",
```

so the example is compliant with what it was described above.

### LINKED ISSUE #343: Broken link to ARCP scheme in spec appendix
**Describe the bug**
This link in the spec is broken
> When parsing such crates it is recommended to use the [Archive and Package (arcp) URI scheme](https://tools.ietf.org/id/draft-soilandreyes-arcp-03.html) to establish a temporary/location-based UUID or hash-based (SHA256) base URI.

**URL**
https://www.researchobject.org/ro-crate/specification/1.1/appendix/relative-uris.html#establishing-a-base-uri-inside-a-zip-file
https://www.researchobject.org/ro-crate/specification/1.2-DRAFT/appendix/relative-uris.html#establishing-a-base-uri-inside-a-zip-file

**Suggested fix**
Should this link to https://pure.manchester.ac.uk/ws/portalfiles/portal/76956641/arcp.html ?

elichad:
fixes #304, fixes #343

### DIFF
--- a/docs/_includes/references.liquid
+++ b/docs/_includes/references.liquid
@@ -86,7 +86,7 @@ and is also rendered into the end of the PDF.
 [Apache Jena]: https://jena.apache.org/
 [Fuseki]: https://jena.apache.org/documentation/fuseki2/
 [RDF 1.1 note on IRIs]: https://www.w3.org/TR/rdf11-concepts/#note-iris
-[ARCP]: https://tools.ietf.org/id/draft-soilandreyes-arcp-03.html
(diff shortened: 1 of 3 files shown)


stain (approved):
Thanks!

PR #363, opened 2024-10-10
added roc-validator

stain:
.. and deprecated ro-crate-validator-py

@simleo may want to check on name to be listed as, I think we should link to pypi.


simleo:
LGTM

elichad (approved):

PR #365, opened 2024-10-24
Remove heuristic for nested ro-crate metadata files

stain:
This algorithm appendum to look for `..../ro-crate-metadata.json` URLs seems unnecessary as such JSON-LDs is not conforming to 1.2, 1.1, 1.0 nor 0.2

### DIFF
--- a/docs/_specification/1.2-DRAFT/appendix/changelog.md
+++ b/docs/_specification/1.2-DRAFT/appendix/changelog.md
@@ -29,6 +29,7 @@ excerpt: List of changes in releases of this specifications
 # APPENDIX: Changelog
 
 * RO-Crate 1.2.0
+  * Clarified that the [RO-Crate Metadata Descriptor MUST have `@id` of `ro-crate-metadata.json`](../root-data-entity#ro-crate-metadata-descriptor) by removing conflicting statements elsewhere in the specification [#365](https://github.com/ResearchObject/ro-crate/pull/365)
   * Updated the Bioschemas namespace for properties from `https://bioschemas.org/ComputationalWorkflow#` to `https://bioschemas.org/properties/`. This change affects only the `input` and `output` properties in the [JSON-LD context](../ro-crate-metadata.json).
   * **Change**: Replaced [name-based algorithm for finding root](../root-data-entity.html#finding-the-root-data-entity)  [#198](https://github.com/ResearchObject/ro-crate/issues/198)
   * Updated [algorithm to always use string filter to find root](../appendix/relative-uris.html#finding-ro-crate-root-in-rdf-triple-stores)  [#189](https://github.com/ResearchObject/ro-crate/issues/189)
--- a/docs/_specification/1.2-DRAFT/root-data-entity.md
+++ b/docs/_specification/1.2-DRAFT/root-data-entity.md
@@ -79,11 +79,6 @@ start with `https://w3id.org/ro/crate/`.
 > The `conformsTo` property MAY be an array, to additionally indicate 
 specializing [RO-Crate profiles](profiles).
 
-If the root data entity `@id` is an absolute URI, the RO-Crate is considered
-web-based: in this case, the metadata descriptor SHOULD also have an absolute
-URI as its `@id`, which MUST have `ro-crate-metadata.json` (or
-`ro-crate-metadata.jsonld` in legacy crates) as its last path segment.
-
 ### Finding the Root Data Entity
 
 In most cases, consumers processing the RO-Crate as a JSON-LD graph can thus
--- a/docs/_specification/1.2-DRAFT/root-data-entity.md
+++ b/docs/_specification/1.2-DRAFT/root-data-entity.md
@@ -110,54 +110,6 @@ metadata_entity = entity_map["ro-crate-metadata.json"]
 root_entity = entity_map[metadata_entity["about"]["@id"]]
 ```
 
-More generally, the metadata id can be a URI whose last path segment is
-`ro-crate-metadata.json`, so the above lookup can fail in non-conforming crates, 
-e.g. a JSON-LD file that has imported an RO-Crate using Linked Data mechanisms.
-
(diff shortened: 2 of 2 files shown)


stain:
I think the fall-back algorithm was added by @simleo in https://github.com/ResearchObject/ro-crate/pull/198 -- so Simone, was there some cases of RO-Crates with absolute URI for their `ro-crate-metadata.json` entity? That would violate MUST in RO-Crate 1.2, 1.1, 1.0, 0.2 but I can see it happening if you have absolutized the RO-Crate metadata file by parsing it as JSON-LD.

simleo:
> I think the fall-back algorithm was added by @simleo in #198 -- so Simone, was there some cases of RO-Crates with absolute URI for their `ro-crate-metadata.json` entity? That would violate MUST in RO-Crate 1.2, 1.1, 1.0, 0.2 but I can see it happening if you have absolutized the RO-Crate metadata file by parsing it as JSON-LD.

Since 1.2, detached RO-Crates can have an absolute URI for their `ro-crate-metadata.json` entity, can't they? What I added to the spec was meant to support such detached crates, and is already supported by ro-crate-py, See [find_root_entity_id](https://github.com/ResearchObject/ro-crate-py/blob/619ba5e3b6c4fdd687b42d33c1819761d8db4503/rocrate/metadata.py#L58).

simleo:
https://www.researchobject.org/ro-crate/specification/1.2-DRAFT/root-data-entity.html says:

> If the root data entity `@id` is an absolute URI, the RO-Crate is considered web-based: in this case, the metadata descriptor SHOULD also have an absolute URI as its `@id`, which MUST have `ro-crate-metadata.json` (or `ro-crate-metadata.jsonld` in legacy crates) as its last path segment.

stain:
> https://www.researchobject.org/ro-crate/specification/1.2-DRAFT/root-data-entity.html says:
> 
> > If the root data entity `@id` is an absolute URI, the RO-Crate is considered web-based: in this case, the metadata descriptor SHOULD also have an absolute URI as its `@id`, which MUST have `ro-crate-metadata.json` (or `ro-crate-metadata.jsonld` in legacy crates) as its last path segment.

Well spotted, and that would be why this algorithm was there. That would however be inconsistent with https://www.researchobject.org/ro-crate/specification/1.2-DRAFT/root-data-entity.html#ro-crate-metadata-descriptor which locks down the name:

> The RO-Crate Metadata Document MUST contain a self-describing RO-Crate Metadata Descriptor with the @id value ro-crate-metadata.json (or ro-crate-metadata.jsonld in legacy crates for RO-Crate 1.0 or older) and @type [CreativeWork](http://schema.org/CreativeWork). This descriptor MUST have an [about](http://schema.org/about) property referencing the Root Data Entity’s @id.

Perhaps a second MUST should be included there on the `@id`?  I think we have to decide one or the other.. also there is `@base` coming into the picture but I assume we're talking in JSON land rather than triple land here.

@ptsefton  you have been making detached crates, what have you named the metadata descriptor? [In Dataverse](https://rdr.kuleuven.be/api/datasets/export?exporter=rocrate_json&persistentId=doi%3A10.48804/N3HWOW) for instance it is just called `"ro-crate-metadata.json"` (but that has not made it a proper detached crate yet)

simleo:
There is another hint at the fact that `ro-crate-metadata.json` can be an arbitrary URI in the [section on detached crates](https://www.researchobject.org/ro-crate/specification/1.2-DRAFT/structure.html#detached-ro-crate):

> [Finding the Root Data Entity](https://www.researchobject.org/ro-crate/specification/1.2-DRAFT/root-data-entity#finding-the-root-data-entity) can be harder for consumers of detached crates, particularly if the platform serving the *RO-Crate Metadata Document* is unable to ensure the URI path ends with `…/ro-crate-metadata.json`.

stain:
Decided 2024-11-14

Delete "If the root data entity @id is an absolute URI, the RO-Crate is considered web-based: in this case, the metadata descriptor SHOULD also have an absolute URI as its @id, which MUST have ro-crate-metadata.json (or ro-crate-metadata.jsonld in legacy crates) as its last path segment"

Add Fallback for "ro-crate-metadata.jsonld" for RO-Crate 1.0

simleo:
What about [Finding RO-Crate Root in RDF triple stores](https://www.researchobject.org/ro-crate/specification/1.2-DRAFT/appendix/relative-uris.html#finding-ro-crate-root-in-rdf-triple-stores)? Should the SPARQL query change? Current:

```sparql
PREFIX schema:  <http://schema.org/>

SELECT ?crate ?metadatafile
WHERE {
  ?crate        a                  schema:Dataset .
  ?metadatafile schema:about       ?crate .
  filter(contains(str(?metadatafile), "ro-crate-metadata.json"))
}
```

simleo:
> What about [Finding RO-Crate Root in RDF triple stores](https://www.researchobject.org/ro-crate/specification/1.2-DRAFT/appendix/relative-uris.html#finding-ro-crate-root-in-rdf-triple-stores)? Should the SPARQL query change? Current:
> 
> ```sparql
> PREFIX schema:  <http://schema.org/>
> 
> SELECT ?crate ?metadatafile
> WHERE {
>   ?crate        a                  schema:Dataset .
>   ?metadatafile schema:about       ?crate .
>   filter(contains(str(?metadatafile), "ro-crate-metadata.json"))
> }
> ```

I think it should *not* change, since once the metadata is parsed as JSON-LD into triples, the metadata descriptor will always have an absolute URI, possibly starting with `file:/`

elichad:
> I think it should not change, since once the metadata is parsed as JSON-LD into triples, the metadata descriptor will always have an absolute URI, possibly starting with file:/

@simleo I agree. I guess the query could find multiple values for `?crate` and `?metadatafile` if there are nested crates, but they should be correctly paired in that case so it doesn't seem like a big problem.

elichad (approved):

PR #374, opened 2024-11-27
Add new contributors to team

elichad:
Latest additions from #1 .

Making a PR rather than committing straight to `main` as I want to check how best to include a group rather than an individual - @stain @ptsefton what do you think about this?


stain:
As agreed in community call 2024-11-28 we won't list organisations as contributors, but the new steering commitee will in January decide how we shall list these (and which). Could you modify to only list people?

PR #392, opened 2025-01-10
Make clarifications in 1.2 text

### LINKED ISSUE #380: Specification: Questions and clarifications for 1.2 text
Questions and suggestions that have come up for me while proofreading the spec. There will probably be more as I continue...

1. In [structure.md](https://github.com/ResearchObject/ro-crate/blob/4fc0b3a18664439639b5b13972f79596e7e7de60/docs/_specification/1.2-DRAFT/structure.md), L201:
    > An RO-Crate [packaged with BagIt](appendix/implementation-notes#adding-ro-crate-to-bagit) may be [referencing external files](appendix/implementation-notes#referencing-external-files) which are not present in the _RO-Crate Root_ hierarchy until the BagIt has been _completed_. This method can be used for files that are large, require authentication or otherwise inconvenient to transfer with the RO-Crate, but which should nevertheless still be considered part of the _payload_.
    
    Not clear to me what "This method" refers to - specifically the BagIt packaging method, or the general guidance on referencing external files?

1. Under [Self-describing and self-contained (Attached RO-Crates)](https://github.com/ResearchObject/ro-crate/blob/main/docs/_specification/1.2-DRAFT/structure.md#self-describing-and-self-contained-attached-ro-crates), the last few paragraphs (from "While RO-Crate is well catered for..." onward) are not incorrect, but seem old/out of place relative to the sections before. Could we add a "Summary" heading or similar just above these?
   * (Noting that [this cross-reference](https://github.com/ResearchObject/ro-crate/blob/4fc0b3a18664439639b5b13972f79596e7e7de60/docs/_specification/1.2-DRAFT/root-data-entity.md#L113) would also need updating if so)

1. [metadata.md](https://github.com/ResearchObject/ro-crate/blob/4fc0b3a18664439639b5b13972f79596e7e7de60/docs/_specification/1.2-DRAFT/metadata.md), L162

    > As of 2024-05-23, the CodeMeta URIs do not resolve correctly, but are used here to match the Codemeta JSON-LD context <https://w3id.org/codemeta/3.0> (issue [#275](https://github.com/ResearchObject/ro-crate/issues/275)). The CodeMeta terms `maintainer` and `funding` are not mapped, as these are already defined by schema.org.
    
    Is this still true? If so, update the date to the (pre-)release date.

1. [root-data-entity.md](https://github.com/ResearchObject/ro-crate/blob/4fc0b3a18664439639b5b13972f79596e7e7de60/docs/_specification/1.2-DRAFT/root-data-entity.md), L39
    > An RO-Crate is described using _JSON-LD_ by an _RO-Crate Metadata Document_. As explained in section [RO-Crate structure](structure) this may be stored in an _RO-Crate Metadata File_. In this section we describe the format of the JSON-LD document.
    
     Suggest this line be moved to the top of the [Metadata](https://github.com/ResearchObject/ro-crate/blob/4fc0b3a18664439639b5b13972f79596e7e7de60/docs/_specification/1.2-DRAFT/metadata.md) section, as it would fit better there.

1. Required Root Data Entity properties have been a point of confusion:
    https://github.com/ResearchObject/ro-crate/blob/4fc0b3a18664439639b5b13972f79596e7e7de60/docs/_specification/1.2-DRAFT/root-data-entity.md#L131-L137
    Suggest changing the first line here to "The _Root Data Entity_ MUST have all of the properties listed below. Each property also has requirements that apply to its value:"

1. [data-entities.md](https://github.com/ResearchObject/ro-crate/blob/main/docs/_specification/1.2-DRAFT/data-entities.md) - the "Encoding file paths" section is duplicated, once under "Core Metadata for Data Entities" (L240) and once under "Web-based Data Entities" (L357). Which copy should be removed?

1. Suggest that [Referencing RO-Crates](https://www.researchobject.org/ro-crate/specification/1.2-DRAFT/data-entities.html#referencing-other-ro-crates) be its own section on the Data Entities page, rather than being nested under "Web-based Data Entities", as it is a long section and includes cases where the nested crates aren't web-based. (Won't affect cross-references as long as the heading names are unchanged.)

1. In the example from [Profiles of nested crates](https://github.com/ResearchObject/ro-crate/blob/main/docs/_specification/1.2-DRAFT/data-entities.md#profiles-of-referenced-crates) the profile is given with a version, not a generic identifier. Is this ok/encouraged? It seems odd for it to be fine for profiles (unless a generic is not available) when we require that the base spec [MUST be versionless](https://github.com/ResearchObject/ro-crate/blob/main/docs/_specification/1.2-DRAFT/data-entities.md#referencing-ro-crates-that-have-a-persistent-identifier) in the same situation.

elichad:
Fixes #380 and points 1 & 2 from #381.

### DIFF
--- a/docs/_specification/1.2-DRAFT/contextual-entities.md
+++ b/docs/_specification/1.2-DRAFT/contextual-entities.md
@@ -117,7 +117,7 @@ An [Organization] SHOULD be the value for the [publisher] property of a [Dataset
 }
 ```
 
-An [Organization] SHOULD also be used for a [Person]'s [affiliation] property.
+An [Organization] SHOULD also be used for a [Person]'s [affiliation] property. 
 
 ```json
 {
@@ -244,7 +244,7 @@ A [data entity](data-entities) MAY provide a published DOI [identifier] that pri
   "@id": "figure.png",
   "@type": ["File", "ImageObject"],
   "name": "XXL-CT-scan of an XXL Tyrannosaurus rex skull",
-  "identifier": "https://doi.org/10.5281/zenodo.3479743",
+  "identifier": {"@id": "https://doi.org/10.5281/zenodo.3479743"},
   "citation": {"@id": "http://ndt.net/?id=19249"},
   "encodingFormat": "image/png"
 },
--- a/docs/_specification/1.2-DRAFT/data-entities.md
+++ b/docs/_specification/1.2-DRAFT/data-entities.md
@@ -387,96 +387,11 @@ Note that if a local file is intended to be packaged within an _Attached RO-Crat
 
 ### Directories on the web; dataset distributions
 
-A _Directory File Entry_ or [Dataset] identifier expressed as an absolute URL on the web can be harder to download than a [File] because it consists of multiple resources. It is RECOMMENDED that such directories have a complete listing of their content in [hasPart], enabling download traversal, or are themselves RO-Crates.
-
-#### Referencing other RO-Crates
-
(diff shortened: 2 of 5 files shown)


stain:
OK - but Pending #390 to avoid merge conflict.

PR #397, opened 2025-01-17
Add RdfCrate to `tools_list`

multimeric:
I'm not 100% sure what values are expected for `level` and `status`, but here is my best guess.


elichad (approved):
Thanks @multimeric !

PR #404, opened 2025-02-18
Make clarifications in 1.2 text 

elichad:
Fixes https://github.com/ResearchObject/ro-crate/issues/380 and points 1 & 2 from https://github.com/ResearchObject/ro-crate/issues/381.

Replaces #392 since some of the changes were made/affected by #390.


ptsefton (approved):
These look fine to me

PR #419, opened 2025-03-27
Add recent publications from ReXAFS and BioHackathon

elichad:


stain:
Also add FDO papers on RO-Crate from https://www.tib-op.org/ojs/index.php/ocp/issue/view/24 -- multiple of them use/describe RO-Crate and Webby FDOs

Issue #400, opened 2025-02-06
Data Entity identification and handling in tools


simleo:
This issue is about the behavior of tools such as [rocrate-validator](https://github.com/crs4/rocrate-validator) and [ro-crate-py](https://github.com/ResearchObject/ro-crate-py) when dealing with data entities, and whether the spec is clear enough in this respect. Their current approach is described in https://github.com/crs4/rocrate-validator/issues/62#issuecomment-2624181022 -- https://github.com/crs4/rocrate-validator/issues/62 was opened by me after discussing the handling of files in https://github.com/nextflow-io/nf-prov/pull/39, which is adding Workflow Run RO-Crate support to Nextflow, but then I discussed it with @kikkomep and stopped thinking that it was a bug in the validator.

In short, given this statement in the spec (it's the same in 1.1 and 1.2-DRAFT):

> Where files and folders are represented as _Data Entities_ in the RO-Crate JSON-LD, these MUST be linked to, either directly or indirectly, from the [Root Data Entity](https://www.researchobject.org/ro-crate/specification/1.1/root-data-entity) using the [hasPart](http://schema.org/hasPart) property.

should a tool consider `File` and `Dataset` entities as data entities and check that they are linked from the Root Data Entity (RDE)'s `hasPart` (this is what the validator does) or should it assume that everything that's linked from the RDE's `hasPart` is a data entity (this is what ro-crate-py does)? Or is it right for different tools with different roles to do different things?

I currently think that the validator is doing the right thing, since it's supposed to check for things like forgetting to link data entities from the RDE's `hasPart`. Regarding ro-crate-py, I'm starting to have doubts: in particular, following the "indirectly" bit above, it implements a recursive walk of `hasPart` properties, and I've recently noticed that this leads to a weird situation where `SoftwareApplication` entities are read as data entities because the `ComputationalWorkflow` is also a `File` and it links to the workflow's tools via `hasPart` as prescribed by the Workflow Run Crate profile (which follows Bioschemas in this respect). Maybe it should only follow `hasPart` from `Dataset` and not `File`?

Should the spec statement cited above be made clearer in order to help implementations? E.g.:

[...]

PR #422, opened 2025-03-31
Add tip that `File` and `Dataset` may appear on non-data entities

elichad:
Related to #400, applies my suggestion from this comment https://github.com/ResearchObject/ro-crate/issues/400#issuecomment-2766759017

Also rephrases a statement that confused me!
> That statement "Files in the RO-Crate Root are not necessarily data entities" needs to be reworded - it invites confusion of files in the real world with entities in the `@graph`.
> https://github.com/ResearchObject/ro-crate/issues/400#issuecomment-2641035619


elichad:
Closing in favour of #426

PR #425, opened 2025-04-03
Fix RDF triple loading example

simleo:
Fixes one example where the root data entity is missing the `hasPart`, making it inconsistent with the triple expansion below.


elichad (approved):
good find, thanks @simleo !

PR #426, opened 2025-04-04
Clarify when File/Dataset entities are and are not data entities

elichad:
In short, `File` and `Dataset` entities are data entities if they use an absolute or relative URI, and not data entities if they use a local identifier starting with `#`.

Fixes(?) #400 

See https://github.com/ResearchObject/ro-crate/issues/400#issuecomment-2779152885 for further explanation

### DIFF
--- a/docs/_specification/1.2-DRAFT/data-entities.md
+++ b/docs/_specification/1.2-DRAFT/data-entities.md
@@ -176,7 +176,7 @@ Additionally, `File` entities SHOULD have:
 
 RO-Crate's `File` is an alias for schema.org type [MediaObject], any of its properties MAY also be used (adding contextual entities as needed).  [Files on the web](#data-entities-in-an-attached-ro-crate-that-are-also-on-the-web) SHOULD also use `identifier`, `url`, `subjectOf`, and/or `mainEntityOfPage`.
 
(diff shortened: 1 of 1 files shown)


docs/_specification/1.2-DRAFT/data-entities.md
```diff
@@ -170,7 +176,7 @@ Additionally, `File` entities SHOULD have:
 
 RO-Crate's `File` is an alias for schema.org type [MediaObject], any of its properties MAY also be used (adding contextual entities as needed).  [Files on the web](#data-entities-in-an-attached-ro-crate-that-are-also-on-the-web) SHOULD also use `identifier`, `url`, `subjectOf`, and/or `mainEntityOfPage`.
 
-
+A `File` entity MAY have an `@id` that is a local identifier beginning with `#`, in which case it is **not** considered to be a Data Entity. This is useful for describing physical files which may not be present, for example if they are expected to be created by running a process. In this case the `localPath` property SHOULD be used to indicate that a `File` may be found at that path in some circumstances.
```
elichad, line 179: Should we clarify that such files (and datasets in the next section) can still be added to `hasPart`? I think that is what we said in our discussion @ptsefton @stain ?

ptsefton: Yes allow them in hasPart

elichad: Done

PR #432, opened 2025-04-30
Set up blog section for website

### LINKED ISSUE #408: Blog section for website
We could use a blog section for the RO-Crate website - to post about development work, new tools/use cases, community announcements, etc
Does not need to have a comments section, so no complex social integration needed, just regular markdown pages

Requirements:

- page that lists all blog posts
- link to that page from menu
- folder for blog posts containing an example/template post
- (optional) special template for blog posts which can list authors, tags, date which are included in the front matter
- (optional) "latest news" block on homepage with the most recent post(s)

There isn't a blog section included in the ELIXIR toolkit theme, but the FAIRDOM website (which uses the theme) has made their own custom one which we could take inspiration from: https://fair-dom.org/news .

Jekyll also has some [built-in support for blog posts](https://jekyllrb.com/docs/posts/).

elichad:
Fixes #408.

This can be home to developer blogs, news and announcements for the RO-Crate community, and so on.


elichad:
feedback from community call:

- add blog link to nav bar
- investigate moving the date to the YAML front matter rather than the file name (or both)
- "Blog" is a more appropriate name than "News"
- investigate having one folder per post and putting any images in the folder rather than `assets/img`

elichad:
This is ready for another review! @a-mile could you take a look when you have time?

One outstanding note: while the website will not display posts dated in the future, those posts will still show up in the RSS feed (`jekyll-feed` doesn't seem to have an option to avoid this). So we should be careful with automations based on that feed to set up scheduled social media posts rather than posting straight away.

elichad:
Approved by @stain on a Zoom call today

PR #460, opened 2025-06-23
RO-Crate 1.2 context alias for 1.2-DRAFT

### LINKED ISSUE #458: Request redirect from `/ro/crate/1.2-DRAFT/context` to `/ro/crate/1.2/context`
We’d like to request a redirect from the following URL:

```
https://w3id.org/ro/crate/1.2-DRAFT/context
```

to

```
https://w3id.org/ro/crate/1.2/context
```

The reason for this is that the `1.2-DRAFT` context URL currently does not resolve to anything, yet it has been used in the wild (perhaps unintentionally) during the draft phase of RO-Crate 1.2. As a result, tools attempting to parse RO-Crates using this draft context URL fail to resolve the context and produce faulty or incomplete triples.

A redirect would help ensure backward compatibility and prevent such failures in RO-Crate consumers that may still refer to the draft context.

stain:
This fixes #458 - rather than hardcoding it on w3c side. Also lets the non-w3id former link work. Note we can't redirect at github pages level with HTML refresh as it won't be respected by JSON-LD parsers.


elichad:
I suggest that we do this within the `_compatibility` folder so we don't accidentally delete the `1.2-DRAFT` folder in future. I've incorporated this fix into #457 as an alternative.

stain:
Agreed, closing as superseded by #457

PR #461, opened 2025-06-24
Add vocabulary guide

### LINKED ISSUE #444: List suggested vocabularies on website
The RO-Crate spec is quite clear about schema.org being the "base" vocabulary, and also suggests a few other vocabularies such as CodeMeta which are in the JSON-LD context and also in the [RO-Crate Metadata](https://www.researchobject.org/ro-crate/specification/1.2-RC1/metadata.html) section of the spec.

However, the world of RDF-compatible ontologies and vocabularies is much wider, and I think it would be helpful for us to point users to some good sources so that they can use this prior work without needing to define things from scratch. Also, I think it would help encourage the adoption of RO-Crate if we can say "look at this plethora of terms you can use in RO-Crate" and "look at these existing standards we are compatible with".

On the website, I don't think this fits into the "Use Cases" page, because I just want to make people aware of certain sources rather than going into detail about how to use these ontologies. The only real criteria I think we need to mandate is the presence of IRIs, because vocabularies without these can't be properly used in RO-Crate.

## Example

At WEHI we are interested in describing medical research data. From reading the spec, I was aware of bioschemas which does help to describe biological entities, but this is not an extensive vocabulary, and it only defines some 20 classes.

Only this week did I stumble upon [BioPortal](https://bioportal.bioontology.org/), which indexes some **17 million classes** that can be used to describe precise assay types, tissues, variants, diseases etc. EBI's [OLS](https://www.ebi.ac.uk/ols4/) captures a similar scope of terms in a readily searchable way.

If this were on the RO-Crate website, I could have found these much earlier.

multimeric:
Designed to close #444.

### DIFF
--- a/docs/pages/resources/vocabularies.md
+++ b/docs/pages/resources/vocabularies.md
@@ -1,0 +1,108 @@
+---
+title: Vocabularies
+sidebar: resources
+---
+
+The RO-Crate spec uses schema.org as the "base" vocabulary for describing entities.
+However, in some cases, your crates can benefit from additional vocabularies which might offer extra precision when describing specific domains.
+
+## Vocabulary Sources
+
+Below are some helpful links that might be helpful in finding a vocabulary or ontology that is RO-Crate compatible.
+Some vocabularies are indexed in multiple databases.
+
+* [W3C](https://www.w3.org/TR/?filter-tr-name=&tags%5B%5D=data) publishes standards, drafts and notes relating to linked data. Some of these are themselves vocabularies:
+    * [`PROV-O`](https://www.w3.org/TR/2013/REC-prov-o-20130430/#prov-o-at-a-glance) which describes provenance
+    * [`time`](https://www.w3.org/TR/owl-time/#summary) for describing temporal concepts
+    * [`csvw`](https://www.w3.org/TR/tabular-data-primer/) describes how to describe CSV files using linked data
+* [DBO (DBpedia Ontology)](https://dbpedia.org/ontology/) is a general purpose ontology for describing Wikipedia data
+* [awesome-ontology](https://github.com/ozekik/awesome-ontology?tab=readme-ov-file#ontologies-and-vocabularies) links to other useful databases and ontologies
+
+### Life Sciences
+
+* [BioPortal](https://bioportal.bioontology.org/) indexes some 17 million classes relating to the life sciences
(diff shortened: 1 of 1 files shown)


stain:
Thanks, @multimeric! Think we should include this, but just to improve navigation in case someone comes here before the spec, would you mind also adding links into https://www.researchobject.org/ro-crate/specification/1.2/profiles.html#extension-vocabularies and https://www.researchobject.org/ro-crate/specification/1.2/appendix/jsonld.html#adding-new-or-ad-hoc-vocabulary-terms at the first paragraphs? Considering this page will be found directly under "Resources" in menu.

I think also we can expand on how a profile should primarily try to use existing schema.org terms, for instance the Workflow Run Crate profile uses additional classes from the Action hierarchy. https://schema.org/docs/schemas.html can be used to find existing types.

elichad:
haven't reviewed the text but adding to Stian's comment - you will also need to add the title and path to https://github.com/ResearchObject/ro-crate/blob/main/docs/_data/sidebars/resources.yml so it will show up in the Resources sidebar

multimeric:
I've linked to the spec's section on extension vocabs. I'm not sure about the "Add local definitions of ad hoc terms" section considering my page is about importing vocabularies and not defining custom terms.

> I think also we can expand on how a profile should primarily try to use existing schema.org terms, for instance the Workflow Run Crate profile uses additional classes from the Action hierarchy.

I also think this is out of scope because the audience of this article is people who need more precise, domain-specific  vocabulary than schema.org. This is motivated by finding [OBI](https://www.ebi.ac.uk/ols4/ontologies/obi) which does a good job of describing our research metadata. I did mention schema.org in the first sentence though.

stain:
>      I also think this is out of scope because the audience of this article is people who need more precise, domain-specific vocabulary than schema.org.

OK, but without that context then it's difficult to place it straight under "Resources".. @elichad  would this also then need #408  blog section to get a natural home?

multimeric:
Did you envisage this as a high-level overview on all types of vocabulary, including the standard schema.org, custom terms and *also* third party ontologies? I'm happy for my work to form just this third category of such a page. However I note that this risks repeating a lot of what is in the spec, which is what I was trying to avoid.

stain:
Yeah, we should only point into the spec, that's what I meant above. not to try to repeat in a big section. Will have a go. 
Thanks for all the effort so far.

stain:
Now live on https://www.researchobject.org/ro-crate/vocabularies - thank you @multimeric!

PR #468, opened 2025-07-17
Fix spec PDF links

elichad:
In the metadata for the specification releases, which shows up on the landing page for each version of the spec, the "PDF" link actually goes to the single-page HTML version. This PR directs them to the PDF version as intended.


stain:
Could go to https://zenodo.org/records/13751027/files/ro-crate-1.2.0.pdf?download=1 and https://zenodo.org/records/13751027/files/ro-crate-1.2.0.html?download=1 instead - then it would count as downloads there on the Zenodo record.

stain (approved):

PR #479, opened 2025-08-14
Add related use cases to blog posts

elichad:
This will allow blog posts to include related use cases in the front matter, and they will be displayed at the end of the blog post.

For example, adding the following front matter to an example post:
```
related_pages:
    use_cases: [workflowhub, digital-twins]
```

Will result in the following rendering:

Screenshot showing related use cases

Inspired by #478


a-mile (approved):
Nice addition to the blogs posts. Tested locally and working as expected.

PR #485, opened 2025-10-07
Blog post: RO-Crate and WorkflowHub at European Galaxy Days 2025

elichad:
Includes:

- blog post
- Galaxy use case page
- small bug fix for displaying related use cases

I have a second post in the works with more info about the Galaxy import/export features, so I didn't include much detail about those here.


stain (approved):
Thanks!

PR #489, opened 2025-10-10
Add nf-prov use case

simleo:
Adds a new use case for [nf-prov](https://github.com/nextflow-io/nf-prov).

@fbartusch could you please check if it looks OK?


stain:
I think it's good so we can merge as-is

PR #502, opened 2025-11-05
Add use-case page for LabID

LauLauThom:
Hi,

As discussed a couple weeks back, here is a proposal to add LabID as an additional use-case.
I tried building locally with Docker but then noticed from the related issue that it's not possible anymore, so I haven't tested it sorry !

PS : I also added an entry to the gitignore to ignore the .DS_Store from mac

Thanks !

### DIFF
--- a/docs/pages/use_cases/labid.md
+++ b/docs/pages/use_cases/labid.md
@@ -1,7 +1,7 @@
 ---
 page_id: labid
 title: Lab Integrated Data (LabID)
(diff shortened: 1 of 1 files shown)


LauLauThom:
Thanks for the feedback, I accepted the suggestions.
@elichad The preview looks great :) !

elichad (approved):
Looks good now :)

PR #518, opened 2026-01-07
Blog post: Highlighting RO-Crate Features in Galaxy

elichad:


elichad:
> (temporarily back dating the post so it shows up) 

this annoys me too, so I've just added some changes in #419 that will let future posts show up if the `future` setting is used (e.g. `bundle exec jekyll serve --future`, or set `future: true` in `_config.yml`)

PR #521, opened 2026-01-08
add note to upgrade guide on changed context mappings

elichad:


elichad:
waiting on merging this one until we decide what we're doing with https://github.com/ResearchObject/ro-crate/issues/529

elichad:
#529 has been discussed but solution won't be finalized for a while, merging this to cover the documentation gap in the meantime

PR #523, opened 2026-01-14
Add custom RSS/Atom feed for blog

elichad:
This is an Atom feed adapted from https://github.com/jekyll/jekyll-feed/blob/master/lib/jekyll-feed/feed.xml (which is the template currently used for the blog through [jekyll-feed](https://github.com/jekyll/jekyll-feed))

The main reason for this is that there is a bug in jekyll-feed that means future posts show up in the RSS feed when they're not supposed to (they are hidden for the default `posts` collection, but we use a non-default collection for our blog). However, jekyll-feed has not been updated in a couple of years and doesn't look like it has an active maintainer, so while I have raised a PR to fix the bug there (https://github.com/jekyll/jekyll-feed/pull/425), I am not optimistic about it being fixed soon. 

Hence, I have adapted their feed template to suit our blog, including hiding future posts. This means that posts will be added to the feed at the same time as they go live on the website.

I don't think that this feed will require a lot of maintenance once set up. It will take over at the same link as the current feed (https://www.researchobject.org/ro-crate/feed/blog.xml).

### DIFF
--- a/docs/_blog/feed_template.xml
+++ b/docs/_blog/feed_template.xml
@@ -4,7 +4,7 @@ permalink: feed/blog.xml
 title: "Research Object Crate (RO-Crate) | Blog"
 ---
 <?xml version="1.0" encoding="utf-8"?>
-<!--This is an Atom feed adapted from https://github.com/jekyll/jekyll-feed/blob/master/lib/jekyll-feed/feed.xml .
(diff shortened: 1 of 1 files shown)


a-mile (approved):
Tested and working as expected

PR #528, opened 2026-02-04
Resolve #527: Changes Place example key to url

pmslavin:
Closes #527 

Updates the examples for the `Place` contextual entity in both the 1.1 and 1.2 specs to use `url`.


elichad (approved):
thanks @pmslavin ! sorry for forgetting about this one

PR #538, opened 2026-05-04
Converting to detached: metadata descriptor must be ro-crate-metadata.json

### LINKED ISSUE #524: Detached crate example in the spec is not valid
The [Converting from Attached to Detached RO-Crate Package](https://www.researchobject.org/ro-crate/specification/1.2/appendix/relative-uris#converting-from-attached-to-detached-ro-crate-package) section takes as example https://about.workflowhub.eu/Workflow-RO-Crate/1.0/ro-crate-metadata.json and shows the following as (abbreviated) example output of the conversion to detached:

https://github.com/ResearchObject/ro-crate/blob/182b892a854f38799956b268511bbf241c0a0775/docs/_specification/1.2/appendix/relative-uris.md?plain=1#L92-L116

The above metadata, however, is not valid, since the metadata descriptor's `@id` is not `ro-crate-metadata.json` but `https://about.workflowhub.eu/Workflow-RO-Crate/1.0/ro-crate-metadata.json`. [The spec for the metadata descriptor](https://www.researchobject.org/ro-crate/specification/1.2/root-data-entity#ro-crate-metadata-descriptor) says that the `@id` must be `ro-crate-metadata.json`, "Even in Detached RO-Crate Packages".

simleo:
Fixes #524.

This PR attempts to resolve #524 by adding instructions to manually change the metadata descriptor's `@id` to `ro-crate-metadata.json` after the conversion.

### DIFF
--- a/docs/_specification/1.3-DRAFT/appendix/relative-uris.md
+++ b/docs/_specification/1.3-DRAFT/appendix/relative-uris.md
@@ -114,7 +114,7 @@ Example output (abbreviated):
 }
 ```
 
-Notice how identifiers like `ro-crate-metadata.json`, `./`, `index.html` and `example/` have been translated to absolute URIs.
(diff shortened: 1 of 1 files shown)


elichad (approved):
thanks @simleo

PR #542, opened 2026-05-20
Update RO-Crate context to use canonical Bioschemas prefix

elichad:
Fixes https://github.com/ResearchObject/ro-crate/issues/529. Opening this PR for discussion with the RO-Crate Steering Committee @ResearchObject/steering-committee - no action to be taken until after that discussion is complete.

I've made these changes to the 1.3-DRAFT version of the spec, as discussed in that issue I think this merits a small 1.3 release.


elichad:
changes discussed and approved by steering committee https://github.com/ResearchObject/governance/issues/20

